PR #108, opened 2015-02-23
Add the scan Wifi implementation

realPy:
Begin decode the NL80211_BSS_ELEMENTS binary elements (rated, channel and ssid)


svinota:
Yeah! Thanks!

PR #120, opened 2015-04-08
Add support for interface based policy routing

welterde:
Just a quick hack to add those in.


svinota:
Merged into a separate branch as #121 

Now we'll have to integrate it and to provide a correct NLA policy

svinota:
Pls revew the last master commits, incl. 0889308728fc1430bd31d7feec87ac198215c59b

variables were renamed:
```
iif -> iifname
oif -> oifname
```
to align the naming scheme with current names used in the kernel code.

Many thanks for the heads-up and for the contribution!

welterde:
Awesome!
Just two things.. you forgot to change the docstring and you still used type instead of action in rule(..).

Thanks!

svinota:
Thanks, fixed.

PR #122, opened 2015-04-21
Return correct errno codes from libc ctypes calls

anguslees:
Previously this code raised OSError exceptions with code errno.ECOMM for
all libc calls via ctypes except the setns syscall where it always used
code '-1'.

This change faithfully propagates the C errno value.


svinota:
Awesome. Thanks a lot!

PR #142, opened 2015-05-19
add ifindex to leave_ibss

creasyw:
This pull request add parameter `ifindex` to the `leave_ibss` so that specific interface is able to leave the mesh network.

### DIFF
--- a/pyroute2/iwutil.py
+++ b/pyroute2/iwutil.py
@@ -364,12 +364,13 @@ class IW(NL80211):
                          msg_type=self.prid,
                          msg_flags=NLM_F_REQUEST | NLM_F_ACK)
 
-    def leave_ibss(self):
+    def leave_ibss(self, ifindex):
         '''
         Leave the IBSS -- the IBSS is determined by the network interface
         '''
         msg = nl80211cmd()
         msg['cmd'] = NL80211_NAMES['NL80211_CMD_LEAVE_IBSS']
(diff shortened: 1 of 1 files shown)


svinota:
Just wondering: is there a reason to make the `index` parameter required? Or it could be optional, as before? Since the first choice changes the API, while the second extends it.

creasyw:
The reason to make this optional `IFINDEX` required is that it has the following error in our OpenWrt system.:
```
>>> from pyroute2 import IW
>>> iw = IW()
>>> iw.leave_ibss()
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.4/site-packages/pyroute2/iwutil.py", line 378, in leave_ibss
    msg_flags=NLM_F_REQUEST | NLM_F_ACK)
  File "/usr/lib/python3.4/site-packages/pyroute2/netlink/nlsocket.py", line 750, in nlm_request
    ret = self.get(msg_seq=msg_seq, terminate=terminate)
  File "/usr/lib/python3.4/site-packages/pyroute2/netlink/nlsocket.py", line 608, in get
    raise msg['header']['error']
pyroute2.netlink.NetlinkError: (22, 'Invalid argument')
```
By adding the `IFINDEX`, it could work as expected. I am not sure about the other develop environment, and this is only based on empirical experience...

svinota:
Ok, merging this — later we can make it optional again, if there will be an issue report.

svinota:
Thanks :)

creasyw:
Thanks a lot:)

Issue #144, opened 2015-05-22
Null terminator for SSID in join_ibss


creasyw:
The Problem is that when I tried to connect the mesh network via `join_ibss`, the SSID will reveal ending with `\x00` at the terminal by using `iw dev`.

The null terminators seem useful for sending some of the messages to netlink in the `pyroute2/netlink/__init__.py`, but this is not the case for the names. So, I come up with a ["hacky" solution](https://github.com/creasyw/pyroute2/commit/2ba8779dc972338229257a2b842e69a9bf489997).

There might be a better solution. If this commit is temporarily acceptable, please let me know how I can improve it, then I will make the pull request.

Thanks a lot!

svinota:
Does the `cdata` type [1] fits this task? Though, the difference will that it requires an external utf-recoding. If it fits, the issue is solved :) if not — let me know.

[1] https://github.com/svinota/pyroute2/blob/0.3.8/pyroute2/netlink/__init__.py#L1273

creasyw:
Thanks! I will try it out, and get back to here tomorrow.

creasyw:
Thanks a lot!! It works! The relevant commit is https://github.com/creasyw/pyroute2/commit/8e8f909ba77f9627522cd510546eaa3606dc1bf9. Should I make the pull request? Or you have some better idea to incorporate this change with some others?

svinota:
If you really need the utf de/encoding in a basic NLA type (that could be useful, I agree), I would rather go on with a new type. Could I ask you to create a new type, let's name it `string` (if it looks reasonable enough), just copy the `cdata` definition, change the name and add encode/decode methods?

The reason why I would leave `cdata` intact is that it should be real `cdata` as CDATA in xml — no de/encoding, just raw binary data.

Thank you a lot.

creasyw:
Sure, of course. My current usage is to send the `ATTR_SSID` in the process of connecting to mesh network. The ending null terminator would introduce problems. But I also observe that the `asciiz` data type is responsible for quite a few useful information rather than only the "names" as plain string.

I will make the pull request, and relate it to this issue.

svinota:
Sorry, I'm on the run now, so missed the pull request :| But the approach is pretty the same, just fewer lines of code :)

Could you pls rebase and re-submit the changeset with SSID type? Thanks and I beg your pardon about this chaos.

creasyw:
Oh, no, not a problem at all!

svinota:
Done

PR #145, opened 2015-05-22
Null terminator for SSID

creasyw:
This pull request solve the problem https://github.com/svinota/pyroute2/issues/144 that the station cannot really connect to mesh network via `join_ibss` because the ending null terminator of `AATR_SSID`. The following code snippet is the output of the `iw dev`.
```
Interface adhoc0
    ifindex 10
    wdev 0x2
    addr aa:dd:db:00:01:a3
    ssid mesh-network\x00
    type IBSS
    channel 36 (5180 MHz), width: 40 MHz, center1: 5190 MHz
```

Note that there is an syntax error when I run the `flake8` in the file that I did not edit...
```
pyroute2/netns/process/base_p3.py:4:37: E901 SyntaxError: invalid syntax 
```

### DIFF
--- a/pyroute2/netlink/__init__.py
+++ b/pyroute2/netlink/__init__.py
@@ -1331,6 +1331,19 @@ class nlmsg_atoms(nlmsg_base):
         '''
         fields = [('value', 's')]
 
+        def encode(self):
+            if isinstance(self['value'], str) and sys.version[0] == '3':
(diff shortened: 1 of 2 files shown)


svinota:
With flake8 it is not an issue — it just two different syntax versions for different Python (2 vs. 3 for the metaclass definition). So the `make test` excludes `base_p2` for Python3 and vice versa.

PR #146, opened 2015-05-22
Change type of ATTR_SSID to string

creasyw:
It is targeting to solve the null terminator at the end of SSID revealing by `iw dev`. https://github.com/svinota/pyroute2/issues/144


svinota:
Thanks!

PR #164, opened 2015-07-24
Set ChanWidth for Join IBSS even if below 40 MHz

ryan3thompson:
Previously we ignored the CHANNEL_WIDTH attribute for all widths under 40 MHz, which defaults them all to 20 MHz No-HT. This pull request will set them to the correct value.


svinota:
Thanks!

PR #166, opened 2015-08-04
Mask SIGINT from netns child processes

drzaeus77:
By default, pressing Ctrl-C on the keyboard when running a python
process in the foreground sends SIGINT to all processes in the
hierarchy, which results in the cleanup+wait of the parent waiting on
some message from the child, which is also cleaning itself up. To avoid
this, mask SIGINT in the children. This places a burden on the parent to
always wait on the child, but avoids the deadlock.

Signed-off-by: Brenden Blanco <bblanco@plumgrid.com>

### DIFF
--- a/pyroute2/netns/nslink.py
+++ b/pyroute2/netns/nslink.py
@@ -68,6 +68,7 @@ import os
 import errno
 import atexit
 import select
+import signal
 import struct
 import threading
 import traceback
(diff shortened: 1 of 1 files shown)


svinota:
Thanks!

drzaeus77:
No problem!

PR #174, opened 2015-08-15
Actually set the FREQ_FIXED flag

ryan3thompson:
This is a flag, so it has a payload of NULL in the netlink `nla_put_flag` command. We previously were not setting this flag, which resulted in off-channel scans despite us setting a fixed frequency.


svinota:
Ok, thanks!

PR #176, opened 2015-08-21
Don't use assertions for flow control

mikkoau:
Converted catching AssertionError to normal if statements. Asserts should not be used for this kind of purpose (program flow control).

We got a tremendous performance improvement (run time dropped from 70 seconds to 3 seconds) after doing this fix on Python 2.x & Py.test based device protocol tester. Py.test does "assertion rewriting" ( http://pybites.blogspot.fi/2011/07/behind-scenes-of-pytests-new-assertion.html ) when an assertion error is encountered in addition to normal call stack tracing. This seems to slow down the program quite a lot.

### DIFF
--- a/pyroute2/netlink/__init__.py
+++ b/pyroute2/netlink/__init__.py
@@ -698,7 +698,8 @@ class nlmsg_base(dict):
         if lvalue is self:
             for key in self:
                 try:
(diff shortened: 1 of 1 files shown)


svinota:
Ough. Thanks

svinota:
And thanks for the hint at py.test, btw.

PR #177, opened 2015-09-03
Add more routing attributes

vodik:
Adds some more recent routing attributes:
    
- `IFLA_PHYS_SWITCH_ID`
- `IFLA_LINK_NETNSID`
- `IFLA_PHYS_PORT_NAME`

All indications show that `IFLA_LINK_NETNSID` is signed but I could be wrong.


svinota:
In the kernel sources there is NLA_S32 policy for `IFLA_LINK_NETNSID`. Sounds strange, but true :)

PR #182, opened 2015-09-16
Remove hanging code from reverted commit

ryan3thompson:
There was an incomplete revert of commits on Aug 11:
https://github.com/svinota/pyroute2/commit/7b27abd65f3a40405f5618370ade113fc2806c6a
https://github.com/svinota/pyroute2/commit/16a8aced82c65b94eb9e15a7d773a3256763d155

We had trouble decoding NL80211_BSS_INFORMATION_ELEMENTS until I removed these two lines and completed the revert. Many of the information elements were incomplete or incorrect - this pull request fixes that.

### DIFF
--- a/pyroute2/netlink/__init__.py
+++ b/pyroute2/netlink/__init__.py
@@ -926,8 +926,6 @@ class nlmsg_base(dict):
             del self['attrs']
         if self['value'] is NotInitialized:
             del self['value']
-        if self.length:
-            self.buf.seek(self.msg_align(self.offset + self.length))
 
     def encode(self):
         '''


svinota:
ough. Thanks for the fix.

svinota:
Since it touches the core, I have to check it first — will apply today a bit later.

svinota:
Yes, you're right. Thanks again.

svinota:
Opened a feature request #183 

I can image the cost of the investigation you done to find the issue. So to avoid such issues in the future I'm going to provide a test module to check the parser against collected binary data — since we can not expect uniform hardware and features on all the platforms used for testing.

As soon as I commit the module, I would ask you to submit the netlink packets data that were parsed incorrectly — instructions will follow.

And pls excuse for this issue. It's my mistake that we still haven't such test module.

svinota:
If it is not a proprietary information, pls provide raw packet data, that was parsed with errors, the guide is here: https://github.com/svinota/pyroute2/tree/master/tests/data

You can also provide test for the data (see `tests/general/test_parser.py`), but that I can do myself as well as soon as the data file will be provided.

Thus we can make sure it will not fail with some next parser patch — functional tests in `test_iwutil.py` still cover not so much, even if there is wifi hardware on the test system.

PR #186, opened 2015-10-07
WIP: Add example tc add-filter for IP & subnet matching on src/dst

### LINKED ISSUE #185: [Docs] Explain filter flags for IP/subnets
The [`IPRouteMixin.tc()` method](../blob/84f99fbbce049517300a08c0389bf27765e1e4f6/pyroute2/iproute.py#L777) has good documentation for most components but is a bit lacking on the bitwise math necessary to create a valid filter.  For example, the following snippet:

> ```py
> # two filters: one to load packets into 1:10 and the
> # second to 1:20
> ip.tc("add-filter", "u32", eth0,
>       parent=0x10000,
>       prio=10,
>       protocol=socket.AF_INET,
>       target=0x10010,
>       keys=["0x0006/0x00ff+8", "0x0000/0xffc0+2"])
> ip.tc("add-filter", "u32", eth0,
>       parent=0x10000,
>       prio=10,
>       protocol=socket.AF_INET,
>       target=0x10020,
>       keys=["0x5/0xf+0", "0x10/0xff+33"])
> ```

It may not be clear to most users how to compose the `keys` argument or even what the example `keys` arguments are describing.

tomislacker:
### Overview
- Add some documentation to for quick-reference on creating `tc` filters for matching on source *or* destination IP/subnet.
- Bit offsets were discovered by using `tc` to add filters via CLI, assessing the output, and re-creating with pyroute2.
- Fixes #185 

### Issues Blocking Merge
1. ~~Resolve merge conflict~~
1. ~~This example is only applicable for interfaces, such as [IMQ]() devices~~

    *@svinota requests this being annotated in the docstring* *(See [a7f83955](https://github.com/tomislacker/pyroute2/commit/a7f83955d6c159785835b3234f7594118273a3cb))*
    
    If you're using classes/qdiscs that aren't available for upstream traffic control, I believe filter additions will be rejected by the kernel that match on destination host and have their `target` set to such a class/qdisc

### DIFF
--- a/pyroute2/iproute.py
+++ b/pyroute2/iproute.py
@@ -792,6 +792,12 @@ class IPRouteMixin(object):
             ff:0   ->   0xff0000
             ffff:1 -> 0xffff0001
 
+        Target notice: if your target is a class/qdisc that applies an
+        algorithm that can only apply to upstream traffic profile, but your
+        keys variable explicitly references a match that is only relevant for
+        upstream traffic, the kernel will reject the filter.  Unless you're
+        dealing with devices like IMQs
+
         For pyroute2 tc() you can use both forms: integer like 0xffff0000
         or string like 'ffff:0000'. By default, handle is 0, so you can add
         simple classless queues w/o need to specify handle. Ingress queue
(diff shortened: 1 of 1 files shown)


svinota:
There's nothing blocking, except it would be cool if you rebase the changeset so there will be no merge conficts.

The issue you described above, with IMQ etc, is better to write in the docstring together with the sample.

svinota:
Thanks!

Issue #187, opened 2015-10-12
[iproute2 rules] Cannot specified a fwmark with mask


tomislacker:
### Observation

If you use the `ip rule` command manually, you can create masked firewall marks:

```sh
ip rule add pref 1000 fwmark 0x100/0x101 lookup mytable
```

```python
from pyroute2.iproute import IPRoute

ip = IPRoute()

fwmark = 0x100
fwmask = 0x101

# Adding as a concat string separated by a slash?
# Nope, the .rule() method will not parse it and an exception is thrown
ip.rule("add",
    priority=1000,
    table=1,
    fwmark="/".join([hex(fwmark), hex(fwmask)]))


# Adding by direct attr injection into method?
# Also no, the .rule() method does not check kwargs for their presence and there
# is no differentiation in the resulting rules from calling with the same arguments
# minus FRA_FWMARK and FRA_FWMASK
ip.rule("add",
    priority=1000,
    table=1,
    FRA_FWMARK=fwmark,
    FRA_FWMASK=fwmask)
```

### Proposition
I think the latter example would probably work out better in that there's less code to be added, less tests, and lower-level control with the addition of a few `if` blocks in [pyroute2.iproute.IPRoute.rule()](../blob/84f99fbbce049517300a08c0389bf27765e1e4f6/pyroute2/iproute.py#L1099).

svinota:
got it, thanks

tomislacker:
I think we can close this is there's no intention to add a *"user-friendly"* kwarg for it or anything which I don't think is entirely necessary

svinota:
I will keep it still, since want to completely re-write kwarg parser in `rule`, so it will accept user-friendly args (as it does now) as well as in the NLA form like `FRA_FWMARK`

svinota:
(excuse me for the rare updates on the project last days, I'm at the OpenStack summit for a while, will be back the next week)

svinota:
Now the `rule()` implementation supports both, «normalized» and «raw» NLA naming. You can provide «fwmask=X» as well as «FRA_FWMASK=X».

Thanks for you patience.

PR #188, opened 2015-10-12
Handling FRA_FWMASK parameter when calling IPRoute.rule()

tomislacker:
This would allow for inserting iproute2 rules with masked `fwmark` parameters as discussed in #187


svinota:
Thanks!

tomislacker:
No problem sir. Love this library and I'm happy I can help add things that
others may use as well.

PR #191, opened 2015-10-18
(partially) repaired inet, repaired icmp6_stats

mtiny:


svinota:
Two questions:
1. Did you check it with flake8 (PEP8 compliance)? Since the CI fails. I can fix it myself, but probably it's better to commit already checked code.
2. Does it work both for 3.x and 4.x kernels?

And thanks for contribution!

mtiny:
I omitted some comment to restore pep8 compatibility. The code works for python 2 and 3 on a 3.16.0-4-amd64 and a 4.1.0-2-686-pae kernel. With inet I am not entirely satisfied. The 'dummy' field is a hack. Otherwise all values will be off by one position. This probably means a bug elsewhere in the program (maybe parsing of netlink message).

svinota:
Ok, thanks. I will try to investigate what's wrong with fields offset, so we'll fix the `dummy` one.

PR #195, opened 2015-11-15
Fix a few typos in the docs

alexwlchan:


svinota:
Thanks so lot!

svinota:
As you see, English is not my strong side :) So your help is important.

PR #206, opened 2015-12-11
make Dotkeys compliant with hasattr protocol

roolebo:
Current implementation of `__getattribute__` in Dotkeys doesn't throw AttributeError if a requested attribute cannot be found. Instead, `KeyError` is thrown and it breaks hasattr as the latter determines non-existence of an attribute by [catching](https://github.com/python/cpython/blob/master/Python/bltinmodule.c#L1051) `AttributeError`.

The change enables rpyc & pyroute2 interopability to manage network on a remote system.

### DIFF
--- a/pyroute2/common.py
+++ b/pyroute2/common.py
@@ -189,8 +189,10 @@ class Dotkeys(dict):
                     self[key[4:]] = value
                     return self
                 return set_value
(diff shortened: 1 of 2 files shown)


svinota:
Thanks! Merged.

PR #220, opened 2016-02-23
Nftables updates

FirefighterBlu3:
this updates the nfnetlink defines to match those in the current linux kernel code (4.4.1). additionally, i put a dotted line between debug prints for clarity.


svinota:
Thanks!

FirefighterBlu3:
i have a starter for nftables, but it looks like nftables is going to be fairly involved. do you have a plan on how it should be organized?

svinota:
I hadn't any particular plans about nftables yet. But roughly speaking all these netlink classes should follow the same simple scheme:
* netlink socket with some default protocol init
* helper methods that perform specific requests

The closest sample is in `ipset.py`:
* `__init__`: sets versions and the policy — so to say which packet types should be parsed with which message classes.
* `request`: simple wrapper around `nlm_request` with some specific protocol setup
* other methods — particular operations on ipsets.

I believe that nftables protocol implementation can work in the same way. Until Friday I will try to make a skeleton class that will perform some simplest operations on nftables.

FirefighterBlu3:
nft seems to make a lot of netlink conversation when doing things so i expect it's going to be fairly involved for every operation. here's a gist of "nft list sets" with one empty table (filter) defined and one empty set defined (ignoring the help table).

https://gist.github.com/FirefighterBlu3/4df572167e0ea2126b63

PR #223, opened 2016-03-03
Add tc clsact and bpf filter direct-action support

drzaeus77:
In linux 4.5, a new clsact qdisc type is added, which is very similar to
ingress but can also apply to egress packets in a lightweight (lockless)
way.

Also improving performance is the direct-action flag for bpf filters, in
which a program directly returns the action (drop, pass, redirect, etc.)
to take in the classifer.


svinota:
Cool! Thousand thanks!

svinota:
Pls notice that the `tcmsg` internals (and, subsequently, `IPRoute.tc()`) are changed, but not documented yet. I plan to complete some docs this weekend.

I really hope it will help to fix the chaos in the `tcmsg` code.

PR #224, opened 2016-03-06
Parse hfsc "d" parameter as time unit

ti-mo:
Just a small fix, I saw most of the parsing logic was already in place. Haven't found any other issues with the HFSC module in master so far! :+1:


svinota:
Cool, thanks!

PR #233, opened 2016-04-18
Add support for vrf interfaces

tmmorin:
(see linux/Documentation/networking/vrf.txt)


svinota:
Thanks, cool!

svinota:
Thomas, would it be better, if the library provided a plugin API for different interface types? So to create one interface would be == to create and register **one** new file with pre-defined API?

tmmorin:
Thanks for merging!

Having a plugin framework for this may be useful.
But there aren't that many types of interface, so I'm not sure.

PR #257, opened 2016-05-29
Fix NetNS leaks on python 3.3-3.5

roolebo:
I hit an issue where a long-running process reached file descriptor limit. The process used NetNS to enter a namespace. Then, once all needed work was done, the namespace was explicitly closed. I've spent some time on the debugging and found that root cause of the issue is [a pipe that is left open after Process.join() (Process.sentinel)](http://bugs.python.org/issue27151).

The PR is a possible workaround for the pipe leak in NetNS.


svinota:
Aha, that's what I missed. Thousand thanks!

PR #267, opened 2016-07-12
Skip tests on the old kernels

roolebo:
I'm trying to make all tests pass on supported versions of python. Some of the tests aren't intended to pass on very old kernels (<4).

I'm about to get all tests passing on Python 2.x. Probably there will be couple more changesets to fix issues on Python 3.x.

py32 tox environment is dropped as pip doesn't support Python 3.0-3.2.


svinota:
Thanks, great!

PR #268, opened 2016-07-13
ipdb: don't sort ipaddrs with empty flags.

roolebo:
Python 3 has stricter comparison rules for None than Python 2 (i.e. None
cannot be compared with None, as TypeError is raised). We cannot rely on
sorted function for keys that might contain None. Therefore we sort only
ip addresses with non-None flags and then append addresses with empty
flags to the end.

### DIFF
--- a/pyroute2/ipdb/interface.py
+++ b/pyroute2/ipdb/interface.py
@@ -781,6 +781,7 @@ class Interface(Transactional):
             # secondaries first.
 
             r_addr = removed['ipaddr']
+
             def get_flags(x):
                 return self['ipaddr'][x]['flags']
             rip_not_sorted = [a for a in r_addr if get_flags(a) is not None]
--- a/pyroute2/ipdb/interface.py
+++ b/pyroute2/ipdb/interface.py
@@ -781,7 +781,8 @@ class Interface(Transactional):
             # secondaries first.
 
             r_addr = removed['ipaddr']
-            get_flags = lambda x: self['ipaddr'][x]['flags']
+            def get_flags(x):
+                return self['ipaddr'][x]['flags']
             rip_not_sorted = [a for a in r_addr if get_flags(a) is not None]
             rip_empty_flags = [a for a in r_addr if get_flags(a) is None]
 
--- a/pyroute2/ipdb/interface.py
+++ b/pyroute2/ipdb/interface.py
@@ -779,9 +779,16 @@ class Interface(Transactional):
             #
             # One simple way to work that around is to remove
             # secondaries first.
-            rip = sorted(removed['ipaddr'],
-                         key=lambda x: self['ipaddr'][x]['flags'],
+
+            r_addr = removed['ipaddr']
+            get_flags = lambda x: self['ipaddr'][x]['flags']
+            rip_not_sorted = [a for a in r_addr if get_flags(a) is not None]
+            rip_empty_flags = [a for a in r_addr if get_flags(a) is None]
+
+            rip = sorted(rip_not_sorted,
+                         key=get_flags,
                          reverse=True)
+            rip += rip_empty_flags
             # 8<--------------------------------------
             for i in rip:
                 # Ignore link-local IPv6 addresses


svinota:
Yep, true. Thanks.

pyroute2/ipdb/interface.py
```diff
@@ -779,9 +779,18 @@ def error(x):
             #
             # One simple way to work that around is to remove
             # secondaries first.
-            rip = sorted(removed['ipaddr'],
-                         key=lambda x: self['ipaddr'][x]['flags'],
```
svinota: I merged it first, but then realized, that it could be solved in another way, as
```python
key=lambda x: self['ipaddr'][x]['flags'] or 0xffff
```

What woud you say?

roolebo: That's much better if 0xffff is invalid value for flags :)
Ср, 13 июля 2016 г. в 23:13, Peter V. Saveliev <notifications@github.com>:

svinota: The max value so far is 0x800 (IFA_F_STABLE_PRIVACY), so 0xffff looks safe. But, from the other side, IFA_FLAGS is uint32, so 0xffffffff would work here better…

Ok then.

svinota: Ough. I just missed one stupid thing.

Look, there is `flags` byte field as well as uint32 `IFA_FLAGS`. Even if the NLA is not present (→ None), the normal field will be an unsigned int in the range 0…255. So where from come addresses with None-flags?

That should be a bug in IPDB then, I believe.

roolebo: Alright, I will look from the side then. I think I've actually hit the bug in another test on python 2.

Sometimes ip addresses are not getting created. CommitException('ipaddr setup error') is raised. That happened because target addresses didn't appear after three attempts to read current ip addressses. I've added a 0.5 second sleep each time this happens and got another CommitException('target flags is not set')
```
  File "/root/pyroute2/.tox/py26/lib/python2.6/site-packages/pyroute2/ipdb/interface.py", line 888, in commit
    transaction._wait_all_targets()
  File "/root/pyroute2/.tox/py26/lib/python2.6/site-packages/pyroute2/ipdb/transactional.py", line 449, in _wait_all_targets
    raise CommitException('target %s is not set' % key)
CommitException: target flags is not set
```

svinota: I think it was not the address flags, but interface flags. Could you pls tell me the test, I'll try to reproduce?

roolebo:
Oh, right. The PR is about IP flags, but another issue is related to interface flags.
```
diff --git a/pyroute2/ipdb/interface.py b/pyroute2/ipdb/interface.py
index 4e61f06..d5cfeb1 100644
--- a/pyroute2/ipdb/interface.py
+++ b/pyroute2/ipdb/interface.py
@@ -837,6 +837,9 @@ class Interface(Transactional):
                     break
                 else:
                     target = set(transaction['ipaddr']) - real
+                    # sometimes netlink doesn't return new ip nor returns EBUSY
+                    # wait a bit in the case
+                    time.sleep(0.5)
             else:
                 raise CommitException('ipaddr setup error')

```

The issue is intermittent:
```
======================================================================
ERROR: test_ipdb.TestCompat.test_global_netns
----------------------------------------------------------------------
  File "/root/pyroute2/.tox/py26/lib/python2.6/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/root/pyroute2/tests/general/test_ipdb.py", line 1184, in test_global_netns
    assert 'x' + ifA not in self.ip.interfaces
  File "/root/pyroute2/.tox/py26/lib/python2.6/site-packages/pyroute2/ipdb/main.py", line 904, in commit
    target.commit(transaction=tx, rollback=rollback)
  File "/root/pyroute2/.tox/py26/lib/python2.6/site-packages/pyroute2/ipdb/interface.py", line 950, in commit
    raise error
CommitException: target flags is not set
```

roolebo:
Peter, an explicit netlink request retrieves flags for an IP, but the ipaddr attribute of an interface doesn't seem to be updated during a commit.
```
diff --git a/tests/general/test_ipdb.py b/tests/general/test_ipdb.py
index af30f24..9dfd2c0 100644
--- a/tests/general/test_ipdb.py
+++ b/tests/general/test_ipdb.py
@@ -140,11 +140,15 @@ class TestExplicit(BasicSetup):
         index = self.ip.interfaces[if1]['index']
         addr = self.ip.nl.get_addr(index=index)[0]
         assert addr['scope'] == 254
+        assert addr['flags'] is not None
+        assert self.ip.interfaces[if1].ipaddr[0]['flags'] is not None
         assert addr.get_attr('IFA_BROADCAST') is None

         index = self.ip.interfaces[if2]['index']
         addr = self.ip.nl.get_addr(index=index)[0]
         assert addr['scope'] == 0
+        assert addr['flags'] is not None
+        assert self.ip.interfaces[if2].ipaddr[0]['flags'] is not None
         assert addr.get_attr('IFA_BROADCAST') == '172.16.103.128'

     def test_addr_loaded(self):
```

roolebo:
I've localized the issue. It happens in `IPDB.update_addr()`. Consider we have one address passed in a list to `update_addr`:
```
[{'attrs': [['IFA_ADDRESS', '172.16.102.1'],
           ['IFA_LOCAL', '172.16.102.1'],
           ['IFA_LABEL', 'prf8514d36'],
           ['IFA_CACHEINFO',
            {'cstamp': 6346600,
             'ifa_prefered': 4294967295,
             'ifa_valid': 4294967295,
             'tstamp': 6346600}]],
 'event': 'RTM_NEWADDR',
 'family': 2,
 'flags': 128,
 'header': {'error': None,
            'flags': 0,
            'length': 76,
            'pid': 14434,
            'sequence_number': 263,
            'type': 20},
 'index': 17750,
 'prefixlen': 24,
 'scope': 254}]
```
The code that makes raw object that's going to be inserted into ipdb, doesn't get value of `flags` key from `addr`, but rather attempts to get IFA_FLAGS attribute, which appeared only in 3.14:
```
            if self.debug:
                raw = addr
            else:
                raw = {'local': addr.get_attr('IFA_LOCAL'),
                       'broadcast': addr.get_attr('IFA_BROADCAST'),
                       'address': addr.get_attr('IFA_ADDRESS'),
                       'flags': addr.get_attr('IFA_FLAGS'),
                       'prefixlen': addr.get('prefixlen')}
```
There's no IFA_FLAGS in the addr, hence we get None flags:
```
{'flags': None, 'prefixlen': 24, 'local': '172.16.102.1', 'address': '172.16.102.1', 'broadcast': None}
```

Kernel headers [prescribe](http://lxr.free-electrons.com/source/include/uapi/linux/if_addr.h?v=3.14#L22) to ignore `ifa_flags` field, if `IFA_FLAGS` attribute is present. 

I'll make another PR with a revert of the PR and a fix.

svinota:
true. Thanks, fixing

svinota:
Thanks for the fix :)

PR #269, opened 2016-07-15
use flags field if IFA_FLAGS attribute is not present in a NLA

roolebo:
Kernel headers [prescribe](http://lxr.free-electrons.com/source/include/uapi/linux/if_addr.h?v=3.14#L22) to ignore `ifa_flags` field, only if `IFA_FLAGS` attribute is present.


svinota:
Ough, you've already fixed that. Thousand thanks!

svinota:
Ported also to the master branch. Thanks again.

PR #273, opened 2016-07-26
Added support for GTP interface

mjaakkol:
Linux introduced GTP interface not that long ago and this enables pyroute2 to support bringing up the interface. This only implements ifinfmsg level support.


svinota:
Kiitoksia paljon! Liitän

mjaakkol:
Np. Keep up the good work!

On Tue, Jul 26, 2016 at 2:21 PM, Peter V. Saveliev <notifications@github.com

PR #277, opened 2016-08-12
ipset: fix parsing of ADT attributes

ffourcot:
ADT messages have the same standard attributes than CADT messages, but
some values overlap (attributes > 16, see IPSET_ATTR_CADT_MAX in
ip_set.h).

This patch split the two kind of messages for the parsing, allowing to
read the real attributes with the list messages. For example,
IPSET_ATTR_BYTES and IPSET_ATTR_PACKETS are now available.


svinota:
Ough, cool! Thanks a lot!

PR #281, opened 2016-08-30
packaging: Support 0.4.x on both EPEL and Fedora

celebdor:
To successfully build on both EPEL and Fedora for Python3 there's some
macros that need to be used. This patch puts them in the specfile and
updates to the latest release.

Signed-off-by: Antoni Segura Puimedon <antonisp@celebdor.com>


celebdor:
For reference: https://fedoraproject.org/wiki/User:Bkabrda/EPEL7_Python3#Specfiles.2C_Macros.2C_Packaging_Process

svinota:
gràcies!

PR #282, opened 2016-08-30
Add some Ipset options

ffourcot:
Add basic support for ipset options (comment, counters, forceadd, maxelem, hashsize).

All features are tested.


svinota:
Thanks so much!

PR #283, opened 2016-09-02
Various ipset features

ffourcot:
* add rename/flush commands
 * add timeout option (for both add/create commands)
 * better error reporting


svinota:
cool. Thanks

PR #287, opened 2016-09-13
Updated message data buffer reference

mjaakkol:
IPQ used debricated msg.buf.getvalue() and this changes it to use msg.data that I believe to be the right way today

### DIFF
--- a/pyroute2/netlink/ipq/__init__.py
+++ b/pyroute2/netlink/ipq/__init__.py
@@ -114,7 +114,7 @@ class IPQSocket(NetlinkSocket):
         msg['header']['type'] = IPQM_MODE
         msg['header']['flags'] = NLM_F_REQUEST
         msg.encode()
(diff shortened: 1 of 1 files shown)


svinota:
Indeed. Thanks!

mjaakkol:
Thanks Peter.

I noticed that the example doesn't work and I also forgot to tweak the
verdict part for the same buffer reference. I'm not sure but I think the
current implementation uses deprecated queue handling but I'm not sure (not
really an expert on netlink protocols). Let me mess around a couple days
with the system and if I get it working, I'll make all the needed changes
and send it back to you.

Mikko

On Wed, Sep 14, 2016 at 3:25 AM, Peter V. Saveliev <notifications@github.com

svinota:
Sure, thanks. I work not so close with IPQ now anyways, so your help is really important.

PR #289, opened 2016-09-16
ipset: support more ipset types

ffourcot:
We already can create ipsets of several types, but that is not enough
to really use them. Indeed, we must support all entries types in the add and
delete commands.

This patch begins the support of other types than IPv4/IPv6 (like iface
and net) and add the support of ipsets with several dimensions
(hash:net,iface for example).

It adds some complexity to parse entries before to add it, since we must
use the good attribute value. Since we are not aware of the ipset type
when we try to insert an entry, the function caller must pass the type
as string (default is "ip", it does not break previous implementation).

We don't support mac address for now, and not protocol (UDP/TCP) options in for ports. For this last one, it looks tricky to implement since it uses IPSET_ATTR_PROTOCOL parameter, already used as recursive parameter for other calls. We are not sure of the best way to handle that.


svinota:
Thanks, merged

Issue #290, opened 2016-09-21
NetNS assumption about /var/run/netns path could be relaxed


salv-orlando:
In some cases (Docker & friends) the network namespace is not bound to an entry in /var/run/netns

The NetNS constructor might allow for specifying an absolute path (e.g.: /proc/xxxx/ns/net) or a fd to that path. This will spare the need for creating a symlink.

This should be a fairly easy patch, even if I haven't yet tried doing that.
Please let me know if there is a workaround that I've not been able to find by myself.

svinota:
One can change `pyroute2.netns.NETNS_RUN_DIR` prior to operate with network namespaces.

But I agree, it may be not the best API, and maybe it would be also not bad to provide explicit path specification in the constructor, as you expected.

salv-orlando:
Indeed changing NETNS_RUN_DIR is the hack I did so far.
However, as I have pyroute2 installed across several nodes, I would have to do some scripting to patch pyroute2 everywhere, which is not really practical.

I agree this is far from being a serious issue. If I have some time today I will put together a patch and create a PR.

svinota:
Pls notice, one can change NETNS_RUN_DIR in the runtime, there is no need to change the module code itself, but as it is a global variable, it affects all the calls:
```python
from pyroute2 import netns
from pyroute2 import NetNS
netns.NETNS_RUN_DIR = "/var/run/my_service"
ns = NetNS("my_netns")
```

salv-orlando:
good point. I think I can do that too.

svinota:
I think, maybe it's a good idea to provide it as an optional argument to all the calls in `pyroute2.netns` and to the `NetNS` constructor, and the argument would default to NETNS_RUN_DIR, if nothing else is specified.

You may play with that or wait for the weekend, I'll do.

svinota:
Closing this ticket as fixed. If there will be issues with the feature, pls reopen it.

PR #292, opened 2016-09-23
NetNS: Allow full namespace path in constructor

salv-orlando:
If the netns argument to the NetNS constructor is the path of an
existing file, that is interpreted as the full namespace path, and
not appended to netns.NETNS_RUN_DIR.

This patch add a setns_from_path function to the netns module.
This function simply opens the namespace, perform the setns syscall,
and returns the file descriptor. In order to avoid code duplication
the setns function now calls setns_from_path

A test has been added to validate creating a NetNS object using the
path of an existing namespace.

Creating or deleting namespaces using a full path has not been
implemented as a part of this patch.

### DIFF
--- a/pyroute2/netns/__init__.py
+++ b/pyroute2/netns/__init__.py
@@ -164,6 +164,15 @@ def remove(netns, libc=None):
     os.unlink(netnspath)
 
 
+def setns_from_path(netnspath, libc=None):
+    nsfd = os.open(netnspath, os.O_RDONLY)
+    libc = libc or ctypes.CDLL('libc.so.6', use_errno=True)
+    ret = libc.syscall(__NR_setns, nsfd, CLONE_NEWNET)
+    if ret != 0:
+        raise OSError(ctypes.get_errno(), 'failed to open netns', netnspath)
+    return nsfd
+
+
 def setns(netns, flags=os.O_CREAT, libc=None):
     '''
     Set netns for the current process.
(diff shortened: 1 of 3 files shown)


salv-orlando:
fixes #290 (or at least it seems like it fixes it)

@svinota I am not sure if you like this approach or not - please let me know, I'm happy to address any change you'd like to make. I chose to do it in this way as it appears to be the least invasive. I hope you do not mind the refactoring I did in netns/__init__.py and tests/general/test_netns.py

svinota:
Please review the API in the `custom_netns_path` branch, if this approach will work for you. If it will, then I'd like to continue in that way — so all the netns function will act in the same manner.

salv-orlando:
Thanks for looking at the code!

I agree that a proper patch should have modified the behaviour for all ns operations. I did lazily take care only of setns (as mentioned in the commit message) - mostly because all I wanted to was to open a namespace in /proc/<pid>/ns/net

I reviewed your branch and it achieves exactly what I needed. The code is also cleaner as it does not have any special casing in NetNS. I really appreciate you taking time to address my issue.

Feel free to close this PR and open a new one. Or merge your branch with this one so you can leverage the unit test (I actually spent 15 minutes for the patch and 45 for the test...) - whatever you prefer.

svinota:
@salv-orlando , pls wait until tomorrow, there are still issues, but I will fix them in the morning.

And thanks for the code, it helped much :) I'm preparing also a merge commit to merge the test.

svinota:
@salv-orlando ok, looks like working. Thanks :)

salv-orlando:
Thanks a lot for your assistancd!

PR #295, opened 2016-09-29
The ac_uid field is aligned on an 8-byte boundary: pad to suit it

hashbrowncipher:
Python's built in padding doesn't work in this case because we have a
4-byte field with __attribute__((aligned(8))) specified on it.


svinota:
Thanks, merged

PR #296, opened 2016-09-29
Added support for ipset.test

fvichot:
Just a simple patch to add support for "test" in ipsets.

If no error, the entry exists in the set, otherwise a exception is raised with code 4103 (corresponding to "already exists" for add/delete, but means "wasn't found in the set" for test)


svinota:
Thanks! Could you please also fix the code to be PEP8-compatible?

svinota:
Nevermind. Fixed PEP8 and add/delete calls, merged. Thanks.

PR #300, opened 2016-10-10
Sort addresses in IPaddrSet

cunha:
Primary and secondary addresses depend on the order in which they are added to an interface [1] .  This commit makes pyroute2 add IP addresses in the same order Interface.add_ip is called. This is achieved by changing LinkedSet.raw from a plain dict to an OrderedDict and reimplementing set set operators in IPaddrSet. The `reload` function reloads all addresses from the kernel.  This also lets `pyroute2` reload the order primary and secondary addresses on interfaces. (More details in the commit messages.)

 [1] iproute2 docs: http://baturin.org/docs/iproute2/#Notes

### DIFF
--- a/pyroute2/ipdb/interface.py
+++ b/pyroute2/ipdb/interface.py
@@ -18,7 +18,6 @@ from pyroute2.ipdb.transactional import Transactional
 from pyroute2.ipdb.transactional import with_transaction
 from pyroute2.ipdb.transactional import SYNC_TIMEOUT
 from pyroute2.ipdb.linkedset import LinkedSet
-from pyroute2.ipdb.linkedset import IPaddrSet
 from pyroute2.ipdb.exceptions import CreateException
 from pyroute2.ipdb.exceptions import CommitException
 from pyroute2.ipdb.exceptions import PartialCommitException
@@ -120,7 +119,7 @@ class Interface(Transactional):
         with self._direct_state:
             for i in ('change', 'mask'):
                 del self[i]
-            self['ipaddr'] = IPaddrSet()
+            self['ipaddr'] = self.ipdb.init_ipaddr_set()
             self['ports'] = LinkedSet()
             self['vlans'] = LinkedSet()
             self['ipdb_priority'] = 0
@@ -1083,7 +1082,7 @@ class InterfacesDict(TransactionalBase):
             device = self[ifname] = self[index]
 
         if index not in self.ipdb.ipaddr:
-            self.ipdb.ipaddr[index] = IPaddrSet()
+            self.ipdb.ipaddr[index] = self.ipdb.init_ipaddr_set()
 
         if index not in self.ipdb.neighbours:
             self.ipdb.neighbours[index] = LinkedSet()
@@ -1147,7 +1146,7 @@ class AddressesDict(dict):
         # Reload addresses from the kernel.
         # (This is a workaround to reorder primary and secondary addresses.)
         for k in self.keys():
-            self[k] = IPaddrSet()
+            self[k] = self.ipdb.init_ipaddr_set()
         for msg in self.ipdb.nl.get_addr():
             self._new(msg)
         for idx in self.keys():
--- a/pyroute2/ipdb/interface.py
+++ b/pyroute2/ipdb/interface.py
@@ -1143,6 +1143,18 @@ class AddressesDict(dict):
         for msg in self.ipdb.nl.get_addr():
             self._new(msg)
 
+    def reload(self):
+        # Reload addresses from the kernel.
+        # (This is a workaround to reorder primary and secondary addresses.)
+        for k in self.keys():
+            self[k] = IPaddrSet()
+        for msg in self.ipdb.nl.get_addr():
+            self._new(msg)
+        for idx in self.keys():
+            iff = self.ipdb.interfaces[idx]
+            with iff._direct_state:
+                iff['ipaddr'] = self[idx]
+
     def _new(self, msg):
         if msg['family'] == socket.AF_INET:
             addr = msg.get_attr('IFA_LOCAL')
--- a/pyroute2/ipdb/linkedset.py
+++ b/pyroute2/ipdb/linkedset.py
@@ -2,6 +2,7 @@
 '''
 import struct
 import threading
+from collections import OrderedDict
 from socket import inet_pton
 from socket import AF_INET
 from socket import AF_INET6
@@ -35,7 +36,7 @@ class LinkedSet(set):
         self.target = threading.Event()
         self.targets = {self.target: _check_default_target}
         self._ct = None
-        self.raw = {}
+        self.raw = OrderedDict()
         self.links = []
         self.exclusive = set()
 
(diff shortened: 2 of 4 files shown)


cunha:
Submitting this more as a request for comments. The code passes all tests in `test_ipdb` and `test_examples`, but my understanding of pyroute2 is limited. Let me know if you'd like changes.

svinota:
Yep, thanks a lot. Give me time till tonight to review.

svinota:
So, pros:
- [x] there are usecases, when it is important to know, which addresses are the primary ones (there may be several subnets on the same interface, thus «addresses», not «address»)
- [x] the patchset provides more natural behaviour

… and cons:
- [ ] in the benchmark it runs 15-20% slower. To run the benchmark:

```
## pls add PYTHONPATH to `env_keep` in the `sudoers` file
$ export PYTHONPATH=`pwd`
$ sudo time -v python benchmark/ipdb.py
```

It comes, among others, from the fact, that the ordered dict affects any `add_ip()` calls, including those, that are called from the system updates (the difference between system and user updates is that the user updates go to an transaction objects, while the system updates are written directly to the interface objects).

I'm thinking on another approach: maybe explicitly mark the primary address upon `add_ip()`, like `add_ip(…, primary=True)`? Another question will be, how to take these «primary» marks into account upon the `commit()` call without significant performance penalties. Or, maybe, provide an API to choose between the ordered and functional-like IP addr syntax, so those, who don't care which address will be the primary, will get no slowdown. Any ideas?

The IPDB is not so rocket fast (to be honest), and any slowdown affects users seriously.

cunha:
Thanks for the comments! I didn't know about the benchmark. I will get more familiar with the benchmark and play around with the code to try to alleviate the slowdown. If we don't get the slowdown to something reasonable (2% or less?), I will get back with ideas (to get your feedback) on how to implement one of your suggestions (I specially like letting the user choose between ordered and unordered addresses, e.g., during IPDB initialization).

[Can work on this over the weekend. Will try to get back to you Saturday/Sunday.]

svinota:
Thanks a lot. I think, it is a really important improvement.

cunha:
Hi, I checked the benchmarks and agree  the modifications to keep addresses sorted increase CPU time. I investigated where the additional CPU time comes from, and found that:

1. The new iterators in `IPaddrSet` are slower, as they are not built-in.
2. `LinkedSet` (and the old `IPaddrSet`) do not keep field `raw` up-to-date when new instances are created from iterators (e.g., in code like `ips_to_remove = set1 - set2`) [A]. One reason for this is that  `__init__` in `LinkedSet` does not call `add`. Given we are using `raw` to keep track of the order or addresses, we need to keep it up to date, which takes extra cycles. (This is why we have `__init__` in `IPaddrSet`.)

To try and make it easy for you to check, I benchmarked only chancing `raw` to `OrderedDict` and found that it has equivalent performance to plain `dict`. I have made a [separate pull request][1] in case you want to test it on your side (I can rebase this pull request on top of the other one if you'd prefer that). Numbers:

 [1]: https://github.com/svinota/pyroute2/pull/303

```
root@desktop:~/pyroute2# for i in $(seq 10) ; do \time -o master.times --append -f "%U %S %E" python benchmark/ipdb.py ; done
root@desktop:~/pyroute2# git checkout linkedset-ordereddict
root@desktop:~/pyroute2# for i in $(seq 10) ; do \time -o ordereddict.times --append -f "%U %S %E" python benchmark/ipdb.py ; done
root@desktop:~/pyroute2# cut -d " " -f 1 master.times  | ../bin/stats --mean
3.07300
root@desktop:~/pyroute2# cut -d " " -f 1 ordereddict.times  | ../bin/stats --mean                                                                                        
3.06700
```


[A] This actually confused me a bit when first looking at the code; I guess it is works fine because when pyroute2 creates new `IPaddrSet`s from set differences, the `raw` dictionary is no longer needed. But this understanding might be wrong.

cunha:
I have not come up with a clean solution to problems 1 and 2 above (I think the cleanest would be to have `OrderedSet` in the standard library, but oh well).  I was thinking about how to change the interface to let users use ordered and unordered addresses (like you suggested).  I think the cleanest API change is to have `IPDB.__init__` receive an extra `sort_addresses=False` parameter. This is compatible with existing code.

Internally, I was thinking about creating an `IPaddrOrderedSet` class that inherits from `IPaddrSet`, moving the slow iterators and `__init__` inside `IPaddrOrderedSet`, then creating a `IPDB.create_ipaddr_set` method to create one of `IPaddrOrderedSet` or `IPaddrSet` depending on how the IPDB was initialized.

Just wanted to check whether you think this is a good approach or whether you have suggestions.

cunha:
I've updated the pull request to implement the approach in the previous post, adding `sort_addresses=False` to `IPDB.__init__` and initializing instances of `IPaddrSet` using a factory method `IPDB.init_ipaddr_set`.  Let me know if you'd like further changes.

svinota:
Tests are still significantly slower :| (`sudo make test module=general:test_ipdb.py` is 180-200 sec vs. 120-130 sec on the master on my laptop)

@cunha , maybe a stupid question, but I have to ask: the order of the addresses per se is meaningless, it is the **adress' role** that we are trying to affect — will it be primary or secondary.

Will it be enough, if we designate an address as a primary in the `add_ip()` calls? Like `ipdb.interfaces.eth0.add_ip('10.0.0.1/24', primary=True)`?

Again, pros:
* we don't touch then any set operations
* it is more both functional and pythonic ways: running an explicit operation instead of implicit choice of the primary by order
* no need to sync the addresses order with the kernel — since the kernel also doesn't care, just uses the first one in the linked list as a primary.

May be I miss something?

cunha:
I've made sort_addresses the default on the tests (maybe I shouldn't have) at test_ipdb.py:85. I expect the new branch to perform similarly to the master branch with sort_addresses = False.

(Will get back to you on the primary=True suggestion shortly.)

svinota:
> I've made sort_addresses the default on the tests

Ough, mea culpa — I missed that, my apologies. You're right, it is better to have the sorting **on** in the test cycle (actually, both variants should be tested, and are tested anyways, but as `sort_addresses` involves more code, the focus should be on it — so, again, you're right)

cunha:
I don't know netlink that well, but the Linux kernel might not let us control what addresses are primary or secondary.

[net/ipv4/devinet.c](https://github.com/torvalds/linux/blob/master/net/ipv4/devinet.c#L440)

``` {.c}
static int __inet_insert_ifa(struct in_ifaddr *ifa, struct nlmsghdr *nlh,
			     u32 portid)
{
	struct in_device *in_dev = ifa->ifa_dev;
	struct in_ifaddr *ifa1, **ifap, **last_primary;

	ASSERT_RTNL();

	if (!ifa->ifa_local) {
		inet_free_ifa(ifa);
		return 0;
	}

	ifa->ifa_flags &= ~IFA_F_SECONDARY;
	last_primary = &in_dev->ifa_list;

	for (ifap = &in_dev->ifa_list; (ifa1 = *ifap) != NULL;
	     ifap = &ifa1->ifa_next) {
		if (!(ifa1->ifa_flags & IFA_F_SECONDARY) &&
		    ifa->ifa_scope <= ifa1->ifa_scope)
			last_primary = &ifa1->ifa_next;
		if (ifa1->ifa_mask == ifa->ifa_mask &&
		    inet_ifa_match(ifa1->ifa_address, ifa)) {
			if (ifa1->ifa_local == ifa->ifa_local) {
				inet_free_ifa(ifa);
				return -EEXIST;
			}
			if (ifa1->ifa_scope != ifa->ifa_scope) {
				inet_free_ifa(ifa);
				return -EINVAL;
			}
			ifa->ifa_flags |= IFA_F_SECONDARY;
		}
	}
```

[Here is a thread about adding such functionality to the kernel.](https://groups.google.com/forum/#!topic/fa.linux.kernel/qiF29jkCeeQ)

If we can't control primary and secondary addresses, then we would need to remove and add addresses so the `primary=True` parameter is honored. I also don't know pyroute2 very well, but adding or removing addresses seems tricky to implement (e.g., might cause unrelated IP addresses to disappear and break connections).

Choosing the set of primary addresses might not cover all use cases. Our use case is one example: We have two addresses X and Y on one interface attached to network X/24. We want devices on network X/24 to receive ICMP replies from X and other remote devices (not in X/24) to receive replies from Y. To do this, we need two primary addresses and need to add Y before X.

[net/ipv4/devinet.c](https://github.com/torvalds/linux/blob/master/net/ipv4/devinet.c#L1196)

``` {.c}
__be32 inet_select_addr(const struct net_device *dev, __be32 dst, int scope)
{
	__be32 addr = 0;
	struct in_device *in_dev;
	struct net *net = dev_net(dev);
	int master_idx;

	rcu_read_lock();
	in_dev = __in_dev_get_rcu(dev);
	if (!in_dev)
		goto no_in_dev;

	for_primary_ifa(in_dev) {
		if (ifa->ifa_scope > scope)
			continue;
		if (!dst || inet_ifa_match(dst, ifa)) {
			addr = ifa->ifa_local;
			break;
		}
		if (!addr)
			addr = ifa->ifa_local;
	} endfor_ifa(in_dev);
```

svinota:
> I also don't know pyroute2 very well, but adding or removing addresses seems tricky to implement (e.g., might cause **unrelated IP addresses to disappear and break connections**).

Yep, you're right.

> Our use case is one example:

Yes, I've got your point. Ok, let's do that way. I'll check all the tests and merge thereafter.

cunha:
Great news!

I ran test_ipdb and only test_snapshots was failing (but it was failing for
the master branch too, so I didn't dig too deeply). Let me know if you find
anything.

svinota:
Finally, merging. Thanks for the effort, and pls excuse me it took so much time.

PR #303, opened 2016-10-16
Keep entry order in LinkedSet

cunha:
This changes LinkedSet.raw from a plain dict to an OrderedDict,
which keeps track of the order in which entries were added or
deleted.


cunha:
Closing because new code in #300 can be used for benchmarking.

PR #305, opened 2016-10-25
support NETLINK_LISTEN_ALL_NSID netlink option

gescheit:
From netlink documentation:
NETLINK_LISTEN_ALL_NSID (since Linux 4.2)
When set, this socket will receive netlink notifications from all network namespaces that have an nsid assigned into the network namespace where the socket has been opened.  The nsid is sent to user space via an ancillary data.


svinota:
Cool. Thanks

PR #320, opened 2016-12-11
Add support for GRE over IPv6

chrisy:
'ip6gre' is implemented in the kernel using the same IFLA_GRE_XXX
fields as GRE over IPv4, but with IPv6 instead of IPv4 addresses
in the 'local' and 'remote' slots.

This patch adds an NLA map for ip6gre with the appropriate support.
Since the field names would conflict with the existing IFLA_GRE
ones, these new items are prefixed IFLA_IP6GRE instead.

I did experiment with using the same, existing, NLA map for both
address families. Unfortunately the 'target' type requires the address
family to be specified in both directions; to and from the kernel.
This can be achieved when creating GRE links; however the only way
I see of reconstructing the addresses from the kernel is to add a
heuristic to the 'target' type to guess the family; this did not
feel like the correct approach.

In the end, the approach in this patch using new IFLA_IP6GRE NLA
entries appears the cleaner of the two.

### DIFF
--- a/pyroute2/ipdb/interface.py
+++ b/pyroute2/ipdb/interface.py
@@ -37,6 +37,7 @@ def _get_data_fields():
                  'tuntap_data',
                  'vxlan_data',
                  'gre_data',
+                 'ip6gre_data',
                  'macvlan_data',
                  'macvtap_data',
                  'ipvlan_data',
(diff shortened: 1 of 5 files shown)


svinota:
Looks reasonable. Thanks!

PR #323, opened 2016-12-23
ipset: support setting counters on ipset add

ffourcot:
Hello,

This patch is perhaps not perfect, since it uses somewhere "bytes" as a local variable. I did this choice because my opinion is that consistency with ipst command line is very important, but I can amend it if you want.


pyroute2/ipset.py
```diff
@@ -158,7 +158,8 @@ def _entry_to_data_attrs(self, entry, etype, family):
         return attrs
 
     def _add_delete_test(self, name, entry, family, cmd, exclusive,
-                         comment=None, timeout=None, etype="ip"):
+                         comment=None, timeout=None, etype="ip",
+                         packets=None, bytes=None):
```
svinota: Personally I prefer `**kwargs` instead of long argument lists, especially in the case of arguments like `type`, `bytes` etc. And later it allows simply to iterate `kwargs` as the names match NLA (after a uniform transformation).

But your code is more readable, so let's leave it as is for now.

svinota:
Thanks!

PR #325, opened 2017-01-03
Ipset revisions

ffourcot:
Hello,

One more pull requests for ipset part. It helps to manage several ipset types with the same ipset attribute, and to check features supported by running kernel.


svinota:
Thanks!

May I add you as a collaborator, so you would put such changes directly? Since I don't work much with ipset, it may be reasonable.

Beside of that, I would assign occasional ipset-related issues to you ;)

ffourcot:
Hello,

Sure, if that can help. If there is one ipset related bug, I'm today probably guilty...

svinota:
Good then. I'm merging other changes — and welcome.

If there are any quick questions, IRC at freenode, `#pyroute2` channel.

PR #341, opened 2017-03-13
Fix tc u32 action stats

etene:
Hello,
I noticed the u32 filter returned the stats2 class instead of an instance for its `TCA_ACT_STATS` member. As a result, it crashed when trying to parse attached action stats.

I fixed it in the code and wrote a test for it. If you want to reproduce, just run the test on a repository that hasn't got the fix yet.


svinota:
Thanks!

PR #344, opened 2017-03-14
tc mirred action

etene:
Hello,
I added the mirred action for tc, and wrote a test for it.
Not being quite sure where to put the test, I added a TestActions class to test_tc.py, tell me if there's anything wrong with the test and I'll fix it.


svinota:
Thanks! Cool

Issue #270, opened 2016-07-20
In python 2.6, a logger doesn't have a default handler


xning:
When importing pyroute2 on CentOS6, will get following errors

No handlers could be found for logger "pyroute2"

I think perhaps we can add one line when importing logging module like this
    import logging
    logging.basicConfig()

svinota:
Not so easy: https://github.com/svinota/pyroute2/issues/255

Thanks for the feedback. The current way is a bit better than the previous one, but not as good as it could be, so it should be fixed also. Give me some time, I think we'll find a solution.

etene:
@svinota don't you think you could close this one now ? My commit won't fix the issue on 2.6 but it will for later versions at least.

Issue #255, opened 2016-05-24
Duplicate log messages


nachtmaar:
Hi, 
while using pyroute2 I recognized a strange behaviour in the python logging module.

As soon as I import pyroute2, I get all log messages twice, but with two different formatters and/or handlers.

See the code below. The first snippet shows the "normal" behaviour, the second the double logging with pyroute2 imported.

```
import logging

#import pyroute2

logger = logging.getLogger("foo")
logger.setLevel(logging.INFO)
formatter = logging.Formatter("%(levelname)s %(name)s %(asctime)s %(module)s: %(message)s")
handler = logging.StreamHandler()
handler.setLevel(logging.INFO)
logger.addHandler(handler)

logger.info("foo")
```

Result:

```
foo
```

```
import logging

import pyroute2

logger = logging.getLogger("foo")
logger.setLevel(logging.INFO)
formatter = logging.Formatter("%(levelname)s %(name)s %(asctime)s %(module)s: %(message)s")
handler = logging.StreamHandler()
handler.setLevel(logging.INFO)
logger.addHandler(handler)

logger.info("foo")
```

Result:

```
foo
INFO:foo:foo
```

svinota:
wow. Looks weird.

Investigating, thanks

svinota:
Looks like the library added a default stream handler by calling `basicConfig()`. And that's really not a proper way.

Submitted a change. Pls check, if that will work for you.

nachtmaar:
Yes, that fixed the problem.

By the way, installing the latest version of pyroute2 fails for me with pip.

```
pip install -e git+https://github.com/svinota/pyroute2.git@203dfd1574727daa9c725d39634e9ffdf2e2c5da#egg=pyroute2

root@f32f9ecd3cac:/tmp/pyroute2# pip install -e git+https://github.com/svinota/pyroute2.git@203dfd1574727daa9c725d39634e9ffdf2e2c5da#egg=pyroute2
Obtaining pyroute2 from git+https://github.com/svinota/pyroute2.git@203dfd1574727daa9c725d39634e9ffdf2e2c5da#egg=pyroute2
  Cloning https://github.com/svinota/pyroute2.git (to 203dfd1574727daa9c725d39634e9ffdf2e2c5da) to ./src/pyroute2
  Could not find a tag or branch '203dfd1574727daa9c725d39634e9ffdf2e2c5da', assuming commit.
[...]

PR #345, opened 2017-03-14
logging: fix spurious warnings when no logging handler is configured

etene:
Hello,
This adds a `logging.NullHandler` handler to the library's top-level logger, as recommended by [the official Python documentation](https://docs.python.org/2/howto/logging.html#library-config) for libraries.

This fixes an issue that had no satisfactory solution until now:
- We either add a regular log handler, which should not be done in a library as it can cause duplicate logs (issue #255)
- Or we don't, but this produces warnings on logging calls when no handler has been configured by the library user (issue #270)

This is not compatible with Python 2.6 but I can amend the pull request if necessary.

### DIFF
--- a/pyroute2/__init__.py
+++ b/pyroute2/__init__.py
@@ -19,6 +19,10 @@ from pyroute2.netlink.exceptions import \
     NetlinkDecodeError
 
 log = logging.getLogger(__name__)
+# Add a NullHandler to the library's top-level logger to avoid complaints
+# on logging calls when no handler is configured.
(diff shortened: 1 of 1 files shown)


svinota:
No, don't amend it. But it would be cool to put this statement under `if sys.version…`, what do you think?

etene:
That seems to be the way it's usually done here, so I'll just do that !

etene:
Done !

svinota:
Thanks!

PR #346, opened 2017-03-15
add the tc connmark action

etene:
Hello,
I added the tc connmark action + a test for it.

I wasn't quite sure about the proper way to handle the padding in `TCA_CONNMARK_PARMS`, this looks fine to me but tell me if there's a better way. 

As for the test, this is a quite recent addition to the kernel as far as I know, so I wrapped it in a `skip_if_not_supported`.

### DIFF
--- a/pyroute2/netlink/rtnl/tcmsg/act_connmark.py
+++ b/pyroute2/netlink/rtnl/tcmsg/act_connmark.py
@@ -1,0 +1,45 @@
+from pyroute2.netlink import nla
+from pyroute2.netlink import NLA_F_NESTED
+from pyroute2.netlink.rtnl.tcmsg.common import tc_actions
+
+"""
+connmark - netfilter connmark retriever action
+see tc-connmark(8)
+
+This filter restores the connection mark into the packet mark.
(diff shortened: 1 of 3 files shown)


pyroute2/netlink/rtnl/tcmsg/act_connmark.py
```diff
@@ -0,0 +1,45 @@
...
+               )
+
+    class tca_connmark_parms(nla):
+        fields = (('index', 'I'),
+                  ('capab', 'I'),
+                  ('action', 'i'),
+                  ('refcnt', 'i'),
+                  ('bindcnt', 'i'),
+                  ('zone', 'H'),
+                  ('__padding', 'H'),  # XXX is there a better way to do this ?
```
svinota: Not yet. And unfortunately I'm not sure there may be a better way — the structure fields alignment sometimes doesn't work as expected neither in `struct`, nor in `ctypes`. So I would leave this as is, with `__padding`. May be stupid, but portable.

svinota:
Cool. Thanks!

PR #355, opened 2017-04-18
Tc mirred action fix

etene:
This fixes an oversight in the mirred action: I forgot to properly set the action type which led to unexpected behavior such as duplicate packets in redirect mode.

I added a few action types to `pyroute2.netlink.rtnl.tcmsg.common.tc_actions` as well because there was a missing action I needed.


svinota:
Thanks!

PR #358, opened 2017-05-08
Try to reduce memory usage in case of messages containing lots of similiar entries

blackwithwhite666:
This patch reduce memory consumption for some workloads like reading neighbours on hosts with big NDP tables (~10k entries):
```
import socket
from pyroute2 import IPRoute
IPRoute().get_neighbours(socket.AF_INET6)
```
Peak memory usage descreased from 70MB to 30MB, see https://gist.github.com/blackwithwhite666/971bf9a9f8ff1905be5c4cc15decb5b2 for details.

Another important change (and this may change API provided by nlmsg_base): parent is weakref by default. Because of it no reference cycles created so there is no need to track those objects by GC (see graphs from previous link).

jfyi: some tests fails on master branch for me, https://gist.github.com/blackwithwhite666/3d6d6ecd8678c3db05b5c487f1b9315c, i don't find what's the problem with them.

### DIFF
--- a/pyroute2/netlink/__init__.py
+++ b/pyroute2/netlink/__init__.py
@@ -386,6 +386,7 @@ Please notice, that NLA list *MUST* be mutable.
 
 '''
 
+import weakref
 import traceback
 import logging
 import struct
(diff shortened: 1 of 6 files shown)


blackwithwhite666:
Seems that broken build has no relation to PR.

svinota:
Brilliant. Thanks!

Looks like it's wokring correctly both with Python2 and Python3, and the remote API is not broken — all netns-related stuff works as before.

Thank you really much.

PR #362, opened 2017-05-19
tc: allow to change or replace a qdisc

vincentbernat:
Note that I am a bit puzzled that the flags are not the same than for `tc`, but this works. In `tc/qdisc.c`, there is:

```c
	if (matches(*argv, "add") == 0)
		return tc_qdisc_modify(RTM_NEWQDISC, NLM_F_EXCL|NLM_F_CREATE, argc-1, argv+1);
	if (matches(*argv, "change") == 0)
		return tc_qdisc_modify(RTM_NEWQDISC, 0, argc-1, argv+1);
	if (matches(*argv, "replace") == 0)
		return tc_qdisc_modify(RTM_NEWQDISC, NLM_F_CREATE|NLM_F_REPLACE, argc-1, argv+1);
```

### DIFF
--- a/pyroute2/iproute.py
+++ b/pyroute2/iproute.py
@@ -1437,6 +1437,8 @@ class IPRouteMixin(object):
                     'del': (RTM_DELQDISC, flags_make),
                     'remove': (RTM_DELQDISC, flags_make),
                     'delete': (RTM_DELQDISC, flags_make),
+                    'change': (RTM_NEWQDISC, flags_change),
+                    'replace': (RTM_NEWQDISC, flags_replace),
                     'add-class': (RTM_NEWTCLASS, flags_make),
                     'del-class': (RTM_DELTCLASS, flags_make),
                     'change-class': (RTM_NEWTCLASS, flags_change),


svinota:
Actually, I do not remember the reason why it's not as in iproute2. And it looks bad. And probably we have to change this as well, but how to solve that, not breaking the code that already works?

If you have any ideas, pls feel free to propose. Meanwhile, I'm merging the patch, thanks a lot.

vincentbernat:
After reading kernel source, it seems that if you use `NLM_REPLACE` without `NLM_F_CREATE`, it will replace the existing node and error if it doesn't exist (with `NLM_F_CREATE`, it would create it). So, semantically, this is like a change. Without doing anything, it will operate a change of the qdisc which is not something supported by any qdisc. For example, the `sfq` qdisc doesn't support that.

So, I would say that the current implementation is more generic but the "change" is more global as if you stop to specifiy an option, it will be reset to the default configuration. I think we should just keep doing that.

svinota:
Ok than. If you will find a minute to reflect this in the docstring, I would be really thankful.

PR #363, opened 2017-05-19
tc: explain the semantics of "change"

vincentbernat:
They differ from the semantics of "change" from the "tc" tool.


svinota:
Thanks!

Issue #352, opened 2017-03-31
Extending the set of `kind`s is not possible


crosser:
We have a custom kernel driver that creates a new `kind` of netdevice (new `type` in the `ip link add` command). It is controlled over netlink and exposes its own set of attributes.

Command-line `iproute2` tool allows to create a shared object responsible for handling new types of interfaces, and we have such custom shared object, so our netdevice is controllable over standard `ip link` command, that, thanks to the said shared object, recognizes a few new custom keywords specific to our interface.

However, as far as I understand, `pyroute2` does not allow addition of such `kind`-specific modules. Therefore support for a custom `kind` cannot be introduced without modification of the "core" source code of the module.

It would be very useful to be able to write an extension module for a new `kind` of the interface, import it, register it with the core `pyroute2` and use all IPRoute/IPDB features with the custom interface.

svinota:
Yep, you're right. The prototype will be ready the next week, I'll ping you here.

svinota:
Sorry for the delay, it was too complicated change.

The prototype is in the master branch, I think to keep the API — and later extend it (add API versioning, etc). The docs are not ready yet, but you can take one of existing plugins to use as an example.

Again, thanks for the FR.

svinota:
And to add a path with plugins, do before spawning any IPRoute/IPDB instances:
```python
from pyroute2 import config
config.data_plugins_path.append("/path/to/your/own/plugins/")
```

crosser:
Thanks! I am on to try it, expect to provide feedback within a few days.

svinota:
Will you have any questions or proposals — don't hesitate to bring it here. Thanks in advance.

crosser:
What is the correct way to specify a character string of fixed size? Element line this:
```
        [IFLA_MYDEV_DEV]       = { .type = NLA_STRING,
                                   .len = DEVICE_NAME_MAX},
```
in `struct nla_policy`

svinota:
There are three built-in string types in the library so far:
1) `asciiz` — encode → `str + '\0'`; decode → up to `'\0'`
2) `cdata` — take the string as is
3) `string` — UTF-8 encoded

There is no strict length limitation, so if you encode `"bala"`, NLA becomes
```
length = 12  # header + data + zero terminator + alignment
data = "bala\0"
```
or
```
length = 8 # header + data
data = "bala"
```

So if you wanna define NLA:
```python
from pyroute.netlink import nla

class mydev(nla):
[...]

PR #376, opened 2017-06-21
use stdlib for plugins

celebdor:
In order to allow for things like Pyinstaller, it is best to use a more
standard approach to plugin support.

Signed-off-by: Antoni Segura Puimedon <antonisp@celebdor.com>

### DIFF
--- a/pyroute2/config/__init__.py
+++ b/pyroute2/config/__init__.py
@@ -23,6 +23,4 @@ machine = platform.machine()
 arch = platform.architecture()[0]
 kernel = LooseVersion(uname[2]).version[:3]
 
-# IFLA_INFO_DATA plugin system prototype
-_absdir = os.path.dirname(os.path.abspath(common.__file__))
-data_plugins_path = [_absdir + '/netlink/rtnl/ifinfmsg/plugins', ]
+data_plugins_pkgs = []
--- a/pyroute2/netlink/rtnl/ifinfmsg/__init__.py
+++ b/pyroute2/netlink/rtnl/ifinfmsg/__init__.py
@@ -1,4 +1,6 @@
+import importlib
 import os
+import pkgutil
 import sys
 import json
 import errno
@@ -21,6 +23,9 @@ from pyroute2.netlink import nlmsg_atoms
 from pyroute2.netlink.rtnl import RTM_VALUES
 from pyroute2.netlink.rtnl.iw_event import iw_event
 from pyroute2.netlink.exceptions import NetlinkError
+from pyroute2.netlink.rtnl.ifinfmsg import plugins
+
+config.data_plugins_pkgs.append(plugins)
 
 
 # it's simpler to double constants here, than to change all the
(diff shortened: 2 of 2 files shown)


celebdor:
In order to test this I have the following:

trial/test.py
`from pyroute2 import IPDB

ipdb = IPDB()
print(ipdb.interfaces)
`
trial/hook-pyroute2.py
`from PyInstaller.utils import hooks

datas = hooks.copy_metadata('pyroute2')
datas += hooks.collect_data_files('pyroute2')
hiddenimports = hooks.collect_submodules('pyroute2')`

Then, with both pyinstaller and pyroute2 (with the patch in this PR) installed I run:

`pyinstaller --additional-hooks-dir=. test.py`

To try it out, just run:
`./dist/test/test`

svinota:
Yep, you're right — that's much better. The only question — is it possible to import plugins from a custom path?

Pls take a look at that: https://github.com/svinota/pyroute2/issues/352

It would be cool either to have it in some way compatible, or possible to migrate to the new scheme with minimal efforts.

celebdor:
@svinota Since I saw it hardcoded I didn't think anybody was modifying the config. Let me try to adapt it, it would be good if you could then try it, since I don't have any other device kind module to try it with.

svinota:
The config is intended to be modified — not by editing, but from within the code :)

And I will write a functional test today, there is no need to have a real devices for that.

Thanks!

crosser:
So if I want to have my plugin module in a custom dir, shall I need to
```
   sys.path.append("/my/custom/dir")
```
prior to instantiating IPRoute/IPDB object? And have the name of the plugin start with specific prefix? Please clarify, and then I will try it and report.

svinota:
Say you have:
```
/usr/local/pyroute2/data_plugins/:
    __init__.py
    mykind.py
```

Than it should work like that:

```python
import sys
from pyroute2 import config
from pyroute2 import IPRoute

sys.path.append('/usr/local/pyroute2')
config.data_plugins_pkgs.append('data_plugins')

with IPRoute() as ipr:
    ipr.link("add", kind="mykind", ifname="test", …)
```

That's the goal; and mostly I'm interested if that approach is OK for you. Or we could simplify it a bit, hiding the `sys.path` modification under the hood, say, turning the `config.data_plugins_pkgs` into some class that implements all the machinery.

svinota:
@crosser yep, I think in the release it may be simply `config.add_data_plugin()` call with the path as the only argument, and the call would hide all the tricks.

So mostly the API change will be from the config variable to the config call.

crosser:
That should work for me (except, will `__init__.py` be necessary? I use `/usr/lib/ip` directly now, and don't want to pollute it any further.)
Let me know when it's ready for testing.

crosser:
By the way, _for my use case_ `sys.path.append()` is perfectly fine, since I need to import my custom module anyway.

svinota:
@crosser I managed to save the old API (c8622877f27842f81a59446c2d437dde34e9d6cb), so your code should work **without** any changes.

Please let me know, if it doesn't.

crosser:
It works, thanks.

While we are around this piece of code, do you think this would be cleaner?
```
diff --git a/pyroute2/netlink/rtnl/ifinfmsg/__init__.py b/pyroute2/netlink/rtnl/ifinfmsg/__init__.py
index 53b5ec6..b219f0b 100644
--- a/pyroute2/netlink/rtnl/ifinfmsg/__init__.py
+++ b/pyroute2/netlink/rtnl/ifinfmsg/__init__.py
@@ -168,21 +168,18 @@ stats_names = ('rx_packets',
 
 def load_plugins_by_path(path):
     plugins = {}
-    cwd = os.getcwd()
-    os.chdir(path)
     files = set([x.split('.')[0] for x in
                  filter(lambda x: x.endswith(('.py', '.pyc', '.pyo')),
-                        os.listdir('.'))
+                        os.listdir(path))
                  if not x.startswith('_')])
+    sys.path.append(path)
     for name in files:
-        sys.path.append(path)
         try:
             module = __import__(name, globals(), locals(), [], 0)
             plugins[name] = getattr(module, name)
         except:
             pass
-        sys.path.pop()
-    os.chdir(cwd)
+    sys.path.pop()
     return plugins
```

svinota:
Yes, sure, it's better. Would you make a PR?

crosser:
done, PR 377

svinota:
@celebdor 0.4.17 is out

eu-smirnov:
Modifying the syspath and/or list of plugins_pkgs doesn't work for me.
The reason is that the list of plugins gets updated only in pyroute2/netlink/rtnl/ifinfmsg/\_\_init\_\_.py:
``` python
190 for pkg in config.data_plugins_pkgs:
191     data_plugins.update(load_plugins_by_pkg(pkg))
192 
193 for path in config.data_plugins_path:
194     data_plugins.update(load_plugins_by_path(path))
```
which is called way before I modify the path. In my case this is the first import of NetlinkError:
```
    from pyroute2 import NetlinkError
  File "/usr/local/lib/python2.7/dist-packages/pyroute2/__init__.py", line 17, in <module>
    from pyroute2.netlink.rtnl.req import (IPRouteRequest,
  File "/usr/local/lib/python2.7/dist-packages/pyroute2/netlink/rtnl/req.py", line 9, in <module>
    from pyroute2.netlink.rtnl.ifinfmsg import ifinfmsg
  File "/usr/local/lib/python2.7/dist-packages/pyroute2/netlink/rtnl/ifinfmsg/__init__.py"
```
The only way to make it work is to update the plugin list explicitly:
```
_plugins = "/usr/lib/ip"
...
config.data_plugins.update(config.load_plugins_by_path(_plugins))
```

svinota:
Yep. The cause is that commit: https://github.com/svinota/pyroute2/commit/85cbd4a1dd2b1fe1870914453fa43506ac8b90c3

Probably, we have to proceed with explicit plugins registration — but I have to update the docs.

crosser:
I can confirm the finding by @eu-smirnov. I did not notice the problem and claimed that it worked (on June 22), as it turned out, because I had lingering `.pyc` of my plugin module in the system path, leftover from earlier tests. My mistake. In fact, the "old API" (c862287) got broken and did not work after this update.

The API change itself is not a problem for me. But it is probably worth documenting the "official way" somewhere.

svinota:
Unfortunately, the previous «dynamic» modules loading was a kind of nasty hack, that made some hidden magic, and personally I'm happy to get rid of it (just try to take alook at the previous «dynamic» version of the root module — https://github.com/svinota/pyroute2/blob/0.4.17/pyroute2/__init__.py)

But I'm sorry I broke the API for the plugins (actually, the dark side of the implicit magic — you never know, when it gets broken). So yes, I'm to document the explicit way.

My apologies again.

PR #377, opened 2017-06-22
plugins: Minor cleanup of plugin loading code

crosser:
Eliminate unnecessary chdir() and only do sys.path.append() once.


svinota:
Thanks!

PR #383, opened 2017-07-07
netns: ppc64 arch reference fixed

harish-24:
With default value 308, power systems fails to create NetNS object. This patch fixes it with proper value for ppc64* arch.

Signed-off-by: Harish <harish@linux.vnet.ibm.com>


svinota:
Thanks!

PR #397, opened 2017-07-25
IPDB: register_callback: check if `mode` is valid

crosser:
In `register_callback()` check if `mode` is one of 'pre' or 'post'
and raise exception if it is not, the same way as it is already done
in `unregister_callback()`.


svinota:
Thanks!

PR #402, opened 2017-07-30
Adding support for the TC Vlan action

maelk:
This adds support for the TC Vlan Action. The parameters for the python tc function are : 

- *v_action* : can be "push", "pop" or "modify"
- *id* : vlan id, integer (can be hexadecimal)
- *protocol* : can be either "802.1Q" or "802.1ad" . Those values reflect the configuration option of the tc-vlan man page
- *priority* : PCP value

The *v_action* is required, and the vlan *id* is required for "push" and "modify" actions. The protocol defaults to "802.1Q", and the priority to 0 (if unspecified).


svinota:
Thanks a lot!

PR #403, opened 2017-07-31
Add prio qdisc

maelk:
Hello,

This pull request adds the TC Prio qdisc. Please comment. It is a copy of sched_pfifo_fast.py, adding the get_parameters to be able to specify the number of bands and the priomap.

Best regards,


svinota:
Maybe just add `get_parameters()` to the pfifo_fast? As that's mostly the same otherwise.

maelk:
Hello,

Thank you for the comment, I have done the modifications. However, the bands and priomap have no effects for pfifo_fast while they should. For prio the qdisc get set properly. I did not find out why the bands and priomap are not set for pfifo_fast. I can create an issue for that if you wish.

svinota:
Nonono, it was just a question, not an instruction. I didn't look deeply into the `/net/sched/sch_prio.c` code, so I'm just not ready to say for sure what's *needed*.

From my point of view, they're maybe not the same, but lookalike in that sense, so *maybe* it would be simpler. But if there is a possibility that the tc init structure will differ for them, then I'm totally wrong.

On the other side, we always can split the module later… Ok, I merge it as is now, later we'll see.

Thanks a lot!

PR #409, opened 2017-08-19
ipdb: Handle IPv6 tables >= 256

### LINKED ISSUE #408: ipdb: Bogus CommitException when adding IPv6 route to table>=256
```
>>> from pyroute2 import IPDB
>>> ipdb = IPDB()
>>> ipdb.routes.add({'dst': '1234::/64', 'oif': 3, 'table': 255}).commit()
{'metrics': {}, 'oif': 3, 'dst_len': 64, 'family': 10, 'proto': 4, 'tos': 0, 'dst': '1234::/64', 'pref': '00', 'ipdb_priority': 0, 'priority': 1024, 'flags': 0, 'encap': {}, 'src_len': 0, 'table': 255, 'multipath': (), 'type': 1, 'scope': 0, 'ipdb_scope': 'system'}
>>> ipdb.routes.add({'dst': '1234::/64', 'oif': 3, 'table': 256}).commit()
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python2.7/dist-packages/pyroute2/ipdb/routes.py", line 515, in commit
    raise error
pyroute2.ipdb.exceptions.CommitException: target family is not set
```

The route is actually added to the kernel properly despite the exception.

PaulSD:
Fixes #408

### DIFF
--- a/pyroute2/ipdb/routes.py
+++ b/pyroute2/ipdb/routes.py
@@ -1060,9 +1060,7 @@ class RoutingTableSet(object):
         if msg['family'] == AF_MPLS:
             table = 'mpls'
         else:
-            table = msg['table']
-            if table == 252:
-                table = msg.get_attr('RTA_TABLE')
+            table = msg.get_attr('RTA_TABLE', msg['table'])
 
         if table in self.ignore_rtables:
             return


svinota:
I'm not sure, but old kernel versions — IIRC — don't always provide RTA_TABLE NLA, thus comes this check. Let me test a bit before merging.

PaulSD:
This patch should handle that case... If RTA_TABLE is missing, then msg['table'] (rtm_table) is used instead.

The only change here is that I check for RTA_TABLE before looking at rtm_table and only use rtm_table if RTA_TABLE is missing, rather than checking rtm_table for a magic value and using RTA_TABLE only if rtm_table contains that magic value.

svinota:
Indeed. Thanks!

Issue #310, opened 2016-11-15
Original route not removed when modifying priority


svvitale:
I'm attempting to shuffle around the priority on some existing routes.  When I change the priority on a route and commit the changes, I see a new route added with the new priority, but the old one is not removed.  The condition is being detected, but not fully handled (see: https://github.com/svinota/pyroute2/blob/master/pyroute2/ipdb/route.py#L405).  It seems like either the old route entry should be fully removed from the kernel routing table, or it should remain and require the caller to remove it.  In its current form, the ipdb.routes list is left in an incorrect state after the transaction completes.

svinota:
Yep, it's a design issue. Thanks, will try to fix.

svvitale:
I would have submitted a pull request, but I didn't feel confident enough about how the multi-phased transactions operate.  Would it work to simply call the delete method with the snapshot data after the new route is created?

```
# create watchdog
wd = self.ipdb.watchdog('RTM_DELROUTE', **self.wd_key(snapshot))
for route in self.nl.route('delete', **snapshot):
    self.ipdb.routes.load_netlink(route)
wd.wait()
```

svinota:
Should be fixed now.

svvitale:
This appears to work correctly for default routes, but fails to set priority for non-default routes.  It raises a NetlinkError: (3, 'No such process').  Let me know if I can provide additional debug information.

svinota:
Ough. Something I didn't expect. Thanks.

Could you pls provide a sample that causes the issue?

svvitale:
Code:
```python
from pyroute2 import IPDB

with IPDB() as ipdb:
    for idx, route in enumerate(ipdb.routes):
        if route['dst'] != 'default':
            print(route)
            route.priority = 500 + idx
            route.commit()
```

Starting route table:
```
[nodepy] pi@node6:~/nodepy (master) $ ip route
default via 192.168.1.1 dev wlan0  metric 303 
default via 192.168.1.1 dev wlan1  metric 304 
192.168.1.0/24 dev wlan0  proto kernel  scope link  src 192.168.1.173  metric 303 
192.168.1.0/24 dev wlan1  proto kernel  scope link  src 192.168.1.137  metric 304 
[nodepy] pi@node6:~/nodepy (master) $
```

Execution fails on the first route update:
```
[nodepy] pi@node6:~/nodepy (master) $ sudo ~/virtualenv/nodepy/bin/python test.py
[...]

PR #412, opened 2017-08-20
ipdb: match RouteKey to kernel and improve route change handling on key change

PaulSD:
This solves three problems:
1. If the 'tos' field on an existing IPv4 route is changed using pyroute2, then pyroute2 currently adds a new route and fails to remove the old route.  This is fixed by adding the 'tos' to the RouteKey when appropriate (for IPv4 but not IPv6).
2. If the 'scope' or 'oif' field on an existing route is changed, then pyroute2 unnecessarily removes and re-adds the route, causing a brief outage.  This is fixed by removing these fields from the RouteKey so that the route will be updated in-place.
3. If the 'priority' field on an existing route is changed, then pyroute2 removes the old route before adding the new route, causing a brief outage.  This is fixed by adjusting the order of operations so that the new route is added before the old route is removed.

### DIFF
--- a/pyroute2/ipdb/routes.py
+++ b/pyroute2/ipdb/routes.py
@@ -134,6 +134,13 @@ class WatchdogKey(dict):
 
 
 # Universal route key
+# Holds the fields that the kernel uses to uniquely identify routes.
+# IPv4 allows redundant routes with different 'tos' but IPv6 does not,
+# so 'tos' is used for IPv4 but not IPv6.
+# For reference, see fib_table_insert() in
+# https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/net/ipv4/fib_trie.c#n1147
+# and fib6_add_rt2node() in
+# https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/net/ipv6/ip6_fib.c#n765
 RouteKey = namedtuple('RouteKey',
                       ('dst',
                        'table',
@@ -619,7 +626,7 @@ class Route(BaseRoute):
                     else:
                         v = 'default'
                 elif field == 'tos' and msg.get('family') != AF_INET:
-                    # ignore tos field for non-IPv4 routes,
+                    # ignore tos field for non-IPv6 routes,
                     # as it used as a key only there
                     v = None
                 elif v is None:
@@ -638,7 +645,7 @@ class Route(BaseRoute):
                     else:
                         v = ip
                 elif field == 'tos' and msg.get('family') != AF_INET:
-                    # ignore tos field for non-IPv4 routes,
+                    # ignore tos field for non-IPv6 routes,
                     # as it used as a key only there
                     v = None
                 values.append(v)
@@ -1110,7 +1117,7 @@ class RoutingTableSet(object):
 
     def gc_mark_addr(self, msg):
         ##
-        # Find invalid route records after addr delete
+        # Find invalid IPv4 route records after addr delete
         #
         # Example::
         #   $ sudo ip link add test0 type dummy
@@ -1122,9 +1129,13 @@ class RoutingTableSet(object):
         #
         # The route {'dst': '10.1.2.0/24', 'gateway': '172.18.0.1'}
         # will stay in the routing table being removed from the system.
-        # That's because the kernel doesn't send route updates in that
-        # case, so we have to calculate the update here -- or load all
-        # the routes from scratch. The latter may be far too expensive.
+        # That's because the kernel doesn't send IPv4 route updates in
+        # that case, so we have to calculate the update here -- or load
+        # all the routes from scratch. The latter may be far too
+        # expensive.
+        #
+        # See http://www.spinics.net/lists/netdev/msg254186.html for
+        # background on this kernel behavior.
 
         # Simply ignore secondary addresses, as they don't matter
         if msg['flags'] & IFA_F_SECONDARY:
@@ -1157,7 +1168,8 @@ class RoutingTableSet(object):
                             record['route']._gctime = time.time()
 
         elif family == AF_INET6:
-            # TODO: add IPv6 support
+            # Unlike IPv4, IPv6 route updates are sent after addr
+            # delete, so no need to delete them here.
             pass
         else:
             # ignore not (IPv4 or IPv6)
--- a/pyroute2/ipdb/main.py
+++ b/pyroute2/ipdb/main.py
@@ -441,11 +441,11 @@ of 'prefix/mask' string and the route priority (if any)::
 
     In [1]: ipdb.routes.tables[254].idx.keys()
     Out[1]:
-    [RouteKey(dst='default', scope=0, table=254, family=2, ...),
-     RouteKey(dst='172.17.0.0/16', scope=253, table=254, ...),
-     RouteKey(dst='172.16.254.0/24', scope=253, table=254, ...),
-     RouteKey(dst='192.168.122.0/24', scope=253, table=254, ...),
-     RouteKey(dst='fe80::/64', scope=0, table=254, family=10, ...)]
+    [RouteKey(dst='default', table=254, family=2, ...),
+     RouteKey(dst='172.17.0.0/16', table=254, ...),
+     RouteKey(dst='172.16.254.0/24', table=254, ...),
+     RouteKey(dst='192.168.122.0/24', table=254, ...),
+     RouteKey(dst='fe80::/64', table=254, family=10, ...)]
 
 But a routing table in IPDB allows several variants of the
 route spec. The simplest case is to retrieve a route by


svinota:
Thanks, testing!

pyroute2/ipdb/routes.py
```diff
@@ -134,14 +134,19 @@ def __init__(self, route):
...
+# Holds the fields that the kernel uses to uniquely identify routes.
+# IPv4 allows redundant routes with different 'tos' but IPv6 does not,
+# so 'tos' is used for IPv4 but not IPv6.
+# For reference, see fib_table_insert() in
+# https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/net/ipv4/fib_trie.c#n1147
+# and fib6_add_rt2node() in
+# https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/net/ipv6/ip6_fib.c#n765
 RouteKey = namedtuple('RouteKey',
                       ('dst',
-                       'scope',
```
svinota: so the scope is not used as a key?

… I'll try to find a relevant issue, but I believe it was added here after a bug record.

pyroute2/ipdb/routes.py
```diff
@@ -134,14 +134,19 @@ def __init__(self, route):
...
                        'table',
                        'family',
                        'priority',
-                       'oif'))
-RouteKey._required = 5  # number of required fields (should go first)
+                       'tos'))
```
svinota: Good to know, thanks!

pyroute2/ipdb/routes.py
```diff
@@ -459,6 +461,12 @@ def mpcheck(mpset):
                     if old_key in route_index:
                         del route_index[old_key]
                 self.nl.route(devop, **transaction)
+                # If the key has not changed, then the kernel will
+                # update the existing route.  If the key has changed,
+                # then the kernel will add a new route and the old route
+                # must be deleted.
+                if old_key != new_key and devop == 'set':
+                    self.nl.route('del', **dict(old_key._asdict()))
```
svinota: yep.

pyroute2/ipdb/routes.py
```diff
@@ -608,17 +616,25 @@ def make_key(cls, msg):
...
+                # only IPv4 uses 'tos' as part of the RouteKey
+                if field == 'tos' and \
+                        msg.get('family', None) != AF_INET:
+                  continue
                 v = msg.get_attr(msg.name2nla(field))
                 if field == 'dst':
                     if v is not None:
-                        v = '%s/%s' % (v, msg['%s_len' % field])
-                    elif field == 'dst':
+                        v = '%s/%s' % (v, msg['dst_len'])
```
svinota: yep, that's a leftover of an earlier code. Thanks

pyroute2/ipdb/routes.py
```diff
@@ -852,29 +868,17 @@ def describe(self, target, forward=False):
...
-                except KeyError:
-                    # no luck with keys, fallback to simple search
-                    for key in self.idx:
-                        if key[:r] == target[:r]:
-                            return self.idx[key]
-                    else:
-                        raise
+                # no luck with keys, fallback to simple search
+                for key in self.idx:
+                    if key[:r] == target[:r]:
```
svinota: `r` is not definded.

pyroute2/ipdb/routes.py
```diff
@@ -608,17 +616,25 @@ def make_key(cls, msg):
         values = []
         if isinstance(msg, nlmsg_base):
             for field in RouteKey._fields:
+                # only IPv4 uses 'tos' as part of the RouteKey
+                if field == 'tos' and \
+                        msg.get('family', None) != AF_INET:
+                  continue
```
svinota: indentation is not a multiple of four (E111)

pyroute2/ipdb/routes.py
```diff
@@ -608,17 +616,25 @@ def make_key(cls, msg):
...
                         v = 'default'
                 elif v is None:
                     v = msg.get(field, None)
                 values.append(v)
         elif isinstance(msg, dict):
             for field in RouteKey._fields:
+                # only IPv4 uses 'tos' as part of the RouteKey
+                if field == 'tos' and \
+                        msg.get('family', None) != AF_INET:
+                  continue
```
svinota: indentation is not a multiple of four (E111)

pyroute2/ipdb/routes.py
PaulSD: Correct, scope is not used as a key.  You can verify by looking at the kernel source (https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/net/ipv4/fib_trie.c#n1147), or using iproute2:
# ip route add 123.0.0.0/24 via 192.168.0.1 scope 10
# ip route add 123.0.0.0/24 via 192.168.0.1 priority 100
# ip route add 123.0.0.0/24 via 192.168.0.1 tos 1
# ip route
123.0.0.0/24 tos 0x01 via 192.168.0.1 dev eth0 
123.0.0.0/24 via 192.168.0.1 dev eth0 scope 10
123.0.0.0/24 via 192.168.0.1 dev eth0 metric 100
# ip route add 123.0.0.0/24 via 192.168.0.1 scope 11
RTNETLINK answers: File exists
# ip route del 123.0.0.0/24 via 192.168.0.1 scope 10
# ip route add 123.0.0.0/24 via 192.168.0.1 scope 11
# ip route
123.0.0.0/24 tos 0x01 via 192.168.0.1 dev eth0 
123.0.0.0/24 via 192.168.0.1 dev eth0 scope 11
123.0.0.0/24 via 192.168.0.1 dev eth0 metric 100

pyroute2/ipdb/routes.py
PaulSD: Oops!  Will fix.

pyroute2/ipdb/routes.py
PaulSD: Good catch! Will fix.

pyroute2/ipdb/routes.py
svinota: I use flake8 (https://pypi.python.org/pypi/flake8) to catch such things. Actually, even the flake8 plugin to the vim (https://github.com/nvie/vim-pep8). Much easier when it backs me up.

pyroute2/ipdb/routes.py
svinota: Okay. Thanks!

PaulSD: It looks like the bug you are referring to is #310, and the relevant commit is https://github.com/svinota/pyroute2/commit/ae1eb7e80e7ada65ea98b4b50845246c01883c33.

It looks to me like the important part of this commit (the part which actually fixed the issue with changing priority on non-default routes) was:
```
+        if 'scope' not in spec:
+            spec['scope'] = 0
```

Adding 'scope' to RouteKey was not actually required to fix that issue.

PaulSD:
Updated.

pyroute2/ipdb/routes.py
PaulSD: Thanks!  I'll try that.

svinota:
Now it's breaking the regression tests, but I will fix that tonight and merge therafter. Thanks again!

svinota:
Looks like it completely breaks the CI — almost all the route tests are failing. The proposed logic in the keys lookup doesn't work, neither works the commit change.

I will try to adopt your changes step by step now, but it will take some time.

Thanks for your patience :)

svinota:
I tried to apply the code except of the key logic. We'll see, if that will work.

Pls try also if it fits your usecases.

svinota:
The route outages issue should be partially solved also (not solved for multipath routes)

PaulSD:
Hmmm...  It looks like test case failures were because I was skipping setting tos in RouteKeys for IPv4.  It looks like you've fixed this by setting tos to None in RouteKeys for IPv4.  I'm not sure why I didn't catch this in my testing ... it looks like there are some cases where my code works properly under Python2 but not Python3, so maybe it just happened to work by chance when I tested it?

There are still a few issues to be fixed here:

- Some issues with documentation and comments
- For non-multipath routes, oif should not be part of RouteKey:
```
# ip route add 123.0.0.0/24 dev eth0
# ip route add 123.0.0.0/24 dev eth1
RTNETLINK answers: File exists
```
- As you pointed out, this all doesn't work right with multipath routes

I will push updated commits which fix the merge conflicts and split my changes apart a bit.  Then I'll take a look at multipath and see if I can figure out how to fix that.

svinota:
I'm removing `oif` from the keys and refactoring the keys as you proposed initially.

svinota:
yeah, you're right about multipath.

svinota:
if you could test the recent master — it would be cool. Looks like the code becomes a bit better and faster, thanks to you.

PaulSD:
Excellent!  Non-multipath handling looks good now.  I haven't tested multipath yet, I will test that tomorrow.

PaulSD:
BTW, I've updated this pull request to eliminate all of the changes you've committed.  The only remaining changes I have are documentation and comment fixes:
https://github.com/svinota/pyroute2/pull/412/commits/b69f6e71e07bd4e97f94be6ed89f483ba154bfeb
https://github.com/svinota/pyroute2/pull/412/commits/e01f8ffa15c3020429febe8ff4647afaaa51546b

svinota:
Thanks! That was an important work.

PR #414, opened 2017-08-27
Added bitmap support for ipset structure type

AZaugg:
Added ipset support for settype: bitmap:*

Only tested with bitmap:port

`[root@localhost pyroute2]# git log -1 --pretty=oneline
4d1e25d2947ed98212dcb7ff6bbcf1e17b5a4873 Added bitmap support for ipset structure type
[root@localhost pyroute2]# python
Python 2.7.5 (default, Nov  6 2016, 00:28:07)
[GCC 4.8.5 20150623 (Red Hat 4.8.5-11)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> from ipset import IPSet
>>> a = IPSet()
>>> a.create(name='testbitmap1', stype='bitmap:port', bitmap_range='0-10000')
[root@localhost pyroute2]# ipset list
Name: testbitmap1
Type: bitmap:port
Revision: 1
Header: range 0-10000
Size in memory: 80152
References: 0
Members:
`

### DIFF
--- a/pyroute2/ipset.py
+++ b/pyroute2/ipset.py
@@ -136,7 +136,7 @@ class IPSet(NetlinkSocket):
     def create(self, name, stype='hash:ip', family=socket.AF_INET,
                exclusive=True, counters=False, comment=False,
                maxelem=IPSET_DEFAULT_MAXELEM, forceadd=False,
-               hashsize=None, timeout=None):
+               hashsize=None, timeout=None, bitmap_ports_range=None):
         '''
         Create an ipset `name` of type `stype`, by default
         `hash:ip`.
@@ -150,6 +150,8 @@ class IPSet(NetlinkSocket):
         * forceadd -- you should refer to the ipset manpage
         * hashsize -- size of the hashtable (if any)
         * timeout -- enable and set a default value for entries (if not None)
+        * bitmap_ports_range -- set the specified inclusive portrange for
+                          the bitmap ipset structure (0-65536)
         '''
         excl_flag = NLM_F_EXCL if exclusive else 0
         msg = ipset_msg()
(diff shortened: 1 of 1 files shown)


pyroute2/ipset.py
```diff
@@ -161,12 +163,24 @@ def create(self, name, stype='hash:ip', family=socket.AF_INET,
...
         if hashsize is not None:
             data['attrs'] += [["IPSET_ATTR_HASHSIZE", hashsize]]
         if timeout is not None:
             data['attrs'] += [["IPSET_ATTR_TIMEOUT", timeout]]
+        if bitmap_range is not None and 'bitmap' in stype:
+            # Set the bitmap range A bitmap type of set
+            # can store up to 65536 entries
+            try:
+                f, t = [int(item) for item in bitmap_range.split('-')]
+            except:
```
ffourcot: This except should catch only ValueError.

pyroute2/ipset.py
```diff
@@ -161,12 +163,24 @@ def create(self, name, stype='hash:ip', family=socket.AF_INET,
...
+                raise TypeError('Invalid bitmap_range')
+            data['attrs'] += [['IPSET_ATTR_PORT_FROM', f]]
```
ffourcot: Your patch will work only on bitmap:port ipset, since we will need IPSET_ATTR_IP_FROM and IPSET_ATTR_IP_TO for bitmap:ip and bitmap:ip,mac. You should update your commit message, and probably rename the bitmap_range variable to something like bitmap_ports_range

pyroute2/ipset.py
```diff
@@ -161,12 +163,24 @@ def create(self, name, stype='hash:ip', family=socket.AF_INET,
         if forceadd:
             cadt_flags |= IPSET_FLAG_WITH_FORCEADD
 
+        if 'bitmap' in stype and bitmap_range is None:
```
ffourcot: I think that it's probably a good idea to split stype here and check that first element is equal to "bitmap"

pyroute2/ipset.py
svinota: true

AZaugg:
Addressed your comments, let me know if your happy with it.

Apologies dont understand your statement above.

ffourcot:
My question was about adding entry in one ipset:

 a = IPSet()
 a.create(name='foo', stype='bitmap:port', bitmap_range='0-10000')
 a.add("foo", 10, etype="port")

It's currently not supported to add a range: a.add("foo", "10-100", etype="port"). But after a short look in the code, we don't support etype="port" as well, so you probably never add any entries with pyroute2 (we probably have some work to do in this direction).

Thanks for your patch!

svinota:
@ffourcot if you're going to implement ranges, pls, use not the string format (or maybe not only the string format) to specify the range, but tuple or list:

`a.add("foo", (10, 100), etype="port")`

PR #415, opened 2017-08-31
Add support for adding ports to existing ipset

AZaugg:
[root@localhost pyroute2]# git log -2 --pretty=oneline
b2b7c835e07e1ab818ac0bae3fa9d9c55e02ff09 Added docstring for _entry_to_data_attrs method
561af2db37bbff4c71942517c536af18a555457c Added support for adding type:port to existing ipset
[root@localhost pyroute2]# cat TEST.py
from ipset import IPSet
a = IPSet()
import pprint
a.create(name='testport', bitmap_ports_range='1-10000', stype='bitmap:port')
a.add(name='testport', entry='8080',etype='port')
a.add(name='testport', entry='9000-9005',etype='port')
[root@localhost pyroute2]# python TEST.py
[root@localhost pyroute2]# echo $?
0
[root@localhost pyroute2]# ipset list testport
Name: testport
Type: bitmap:port
Revision: 1
Header: range 1-10000
Size in memory: 80144
References: 0
Members:
8080
9000
9001
9002
9003
9004
9005`

### DIFF
--- a/pyroute2/ipset.py
+++ b/pyroute2/ipset.py
@@ -186,6 +186,13 @@ class IPSet(NetlinkSocket):
                             terminate=_nlmsg_error)
 
     def _entry_to_data_attrs(self, entry, etype, family):
+        """
+        Construct the required attributes for IPSET_ATTR_DATA
+            :param entry: str: Entry needed to be added to the ipset
+            :param etype: str: IPSet type, ip, net, port supported
+            :param family: socket.AF_INET*: address family of
+                           entry
+        """
         attrs = []
         if family is not None:
             if family == socket.AF_INET:
--- a/pyroute2/ipset.py
+++ b/pyroute2/ipset.py
@@ -13,6 +13,7 @@ rename, swap, test...)
 '''
 import errno
 import socket
+from re import search
 from pyroute2.netlink import NLMSG_ERROR
 from pyroute2.netlink import NLM_F_REQUEST
 from pyroute2.netlink import NLM_F_DUMP
@@ -195,6 +196,20 @@ class IPSet(NetlinkSocket):
                 ip_version = None
             else:
                 raise TypeError('unknown family')
+
+        if etype == 'port':
+            if not search(r'(^\d+-\d+$)|(^\d+$)', entry):
+                raise ValueError('Bad port or port range')
+            if '-' in entry:
+                t, f = [int(x) for x in sorted(entry.split('-'))]
+                attrs += [
+                    ['IPSET_ATTR_PORT_FROM', f],
+                    ['IPSET_ATTR_PORT_TO', t]
+                ]
+            else:
+                attrs += [['IPSET_ATTR_PORT_FROM', int(entry)]]
+            return attrs
+
         for e, t in zip(entry.split(','), etype.split(',')):
             if t in ('ip', 'net'):
                 if t == 'net':


pyroute2/ipset.py
```diff
@@ -195,6 +203,20 @@ def _entry_to_data_attrs(self, entry, etype, family):
                 ip_version = None
             else:
                 raise TypeError('unknown family')
+
+        if etype == 'port':
+            if not search(r'(^\d+-\d+$)|(^\d+$)', entry):
```
svinota: is there a good reason to have it as a string parameter, not as int / tuple?

```python
port=10
port=(10, 100)
# ...
if isinstance(port, int):
    # add a port
elif isinstance(port, (tuple, list)):
    # add a range using port[0], port[1]
else:
    raise ValueError('...')
```

That would allow us to skip the `re` import and the string operations. The library is used in embedded environments (yeah, I know, but still), and such things are ctritical sometimes.

AZaugg: I wanted to keep it consistent with the how "entry" was being used, it looks like that its always of type string. I know strict typing is mandatory, but I thought it would be easier to maintain. 

Let me know if you still want me to change it? I am happy to do so.

svinota: Ough. You're right, it's a string there :( I didn't take at that before.

May I ask you to rework your code somehow to avoid `re` anyways (I hope that's possible)?

And I'll open a new discussion meanwhile, since I really don't think that having string here is agood idea.

svinota: @AZaugg pls ignore previous comment

@ffourcot , could you please take a look — if it's possible to avoid using strings here in the port value? Or tell me the reason why not. I would really like to see a discussion about that topic, since you're more into ipset, but the common sense tells me that `int → str → int` transformations are not good.

AZaugg: So a quick look, would you be happy to pass in a new argument in, "port":

` 
def _entry_to_data_attrs(self, entry=None, port=None , etype, family):
    if port:
        # to port related stuff
     elif entry:
          # do entry related stuff

`

svinota: @AZaugg let's wait Florent's comments, he's the maintainer of this code.

E.g., I would be happy if the `entry` will be of the types `int, tuple, list` **if** `etype == "port"`, so neither new function arguments, nor strings for int values. But let's wait a bit.

ffourcot: @svinota entry here can have several types, actually all kind of values that ipset modules can contains. It means:
 * interface names (as string)
 * IP addresses/network (with or without netmask)
 * MAC addresses
 * marks/ports (as int)

And we can mix sometimes both or three of them. We know the real type of each value thanks to "etype" parameter, based on "stype" of ipset (example: one ipset stype="hash:net,iface" will take one entry etype="net,iface").

The code was at beginning only written for ip or networks one single dimensions, so strings was not so curious. When we added several dimensions support, the split/zip made easier (at the beginning only, probably) to understand how to call the function.

In my opinion, it's perfectly fine to use a tuple when there are several values, and avoid cast when not necessary (like "mark" and "port" type). We probably should check "entry" variable type at the beginning and split on comma only when it's a string (and only when etype contains several types)

However, it's begin to be more funky when a type (like port here) can take one or more argument. Entry will be a tuple for two reasons:

 > entry = ("10.10.10.10/32", 10) # entry for net,port ipsets
 > entry = (50, 80) # for a bitmap:port, it's a range of ports
 > entry = ("10.0.0.1/32", (50, 80) # for a bitmap:ip,port ipset (it does not exist today I think, but there are sometimes really curious ipsets)

It will be hard to discriminate between normal tuples (because of several types) and port tuple (or other range tuple, we need a least to support IP ranges). In case of range, I vote for a tuple of tuple, so to change the second example by:

  > entry = ((50, 80),)  # easier to iterate on the list of values. It's of course a range
  > entry = ("192.168.0.1", "192.168.0.254")  # it's for a hash:net,net ipset
  > entry = ((192.168.0.1", "192.168.0.254"), ) # it's for a bitmap:ip ipset (with a range)

So, if it's ok for you, that will be something like:

def _entry_to_data_attrs(self, entry, etype, family):
  attrs = []
  [...}
  types = etype.split(",")
  if len(types) > 1 and isinstance(entry, str):
    entry = entry.split(",")
    assert len(entry) == len(types)
  # back to previous code
  for e, t in zip(entry, types):
    [...]
    if t == "port":
      if isinstance(e, (tuple, list)):
        # do a range
     else:
        # do a single port

ffourcot: I just read in man pages that port can sometimes be a little bit more complicated, for hash:ip,port:

```
ipset create foo hash:ip,port
ipset add foo 192.168.1.1,udp:53
ipset list toto2
  192.168.1.1,udp:53
```

For that case, I have no idea to manage it (note that protocol is optional, one can avoid it)

svinota: Ough, almighty Chaos… Ok, guys, so it's totally up to you how to manage that. I fear, the requirement may place some unnecessary limitations so the decision is up to you.

svinota: But please — the less new external imports, the better. Not strictly, but if possible.

ffourcot: I will have a look next week to try to check every usecase/ipset module. Some tests/documentations on current implementation are mandatory too (my bad), I understand that this can be very confusing.

@AZaugg before to merge or ask you for some amends on your commits, I will try to find something to catch all usecases. Don't hesitate if we have some suggestions.

svinota: But wait. We're not in the CLI.

That's only a proposal and may be it's better to start a new discussion, but shortly:

1. introduce data types: `PortRange`, `IPRange`, etc.
2. then:
```python
entry = (PortRange(22, 80), …)
entry = (PortRange(22, 80, proto='tcp'), …)
```
etc.

Not in that particular pull request.

@ffourcot @AZaugg what would you say?

AZaugg: I was trying to solve my very specific usecase, happy to amend to try better support ipset types holistically. ill take a look next week and see what I can come up with

AZaugg: @ffourcot im looking at supporting port regardless of etypes. Having trouble with constructing the struct.  When accepting, a hash:*port
`struct hash_netport4_elem {
	__be32 ip;
	__be16 port;
	u8 proto;
	u8 cidr:7;
	u8 nomatch:1;
};`

I populate it like, but I raise exceptions because of the attributes passed in with IPSET_ATTR_PROTO
`[['IPSET_ATTR_IP_FROM', {'attrs': [['IPSET_ATTR_IPADDR_IPV4', '192.168.1.3']]}], ['IPSET_ATTR_PORT_FROM', 54], ['IPSET_ATTR_PROTO', {'attrs': [6]}]]
{'attrs': [['IPSET_ATTR_PROTOCOL', 6],
           ['IPSET_ATTR_SETNAME', 'foo'],
           ['IPSET_ATTR_DATA',
            {'attrs': [['IPSET_ATTR_IP_FROM',
                        {'attrs': [['IPSET_ATTR_IPADDR_IPV4',
                                    '192.168.1.3']]}],
                       ['IPSET_ATTR_PORT_FROM', 54],
                       ['IPSET_ATTR_PROTO', {'attrs': [6]}]]}]],
 'header': {},
 'nfgen_family': 0,
 'res_id': 0,
 'value': <class 'pyroute2.netlink.NotInitialized'>,
 'version': 0}`

Are you able to help me out, inregards to constructing the attrs for  IPSET_ATTR_PROTO, I thought it would be:
`#define IPPROTO_IPIP		IPPROTO_IPIP
  IPPROTO_TCP = 6,		/* Transmission Control Protocol	*/`

AZaugg: In addition, I think we should leave the type as a string, this way we can reuse the loop for all etypes.

PR #420, opened 2017-09-15
Add new wiset module

ffourcot:
IPSet object from ipset.py is providing a low level API to manage
NFNetlink sockets. It's "stateless" since it does not have any context
of the ipset manipulated (for example, to know ipset entries and help to
parse responses). Moreover, it's made to manipulate all ipsets in the
same time.

This module provides or more high level API, based on WiSet object. A
WiSet is directly mapped to a real ipset object in kernel, and help to
store metadata like:
 - type of entries
 - configuration of ipset (counters enabled, comments, etc)
 - low level configuration like hash table size, module revision, etc

All data are stored in native python format, and you don't need to know
IPSET_ATTR_* variable to set or read values. Netlink interface is fully
abstract with this object.

It's help to run a batch of command on a set without to have to provide
at each call all relevant metadata. It also provides helpers like
"replace_entries" method, to fill a set with a new content is an atomic
call.

Since some magic is made to abstract netlink socket provided by IPSet
object, we provide several way to make life of users easier:

 - you never need to explicit open/close of sockets if you don't want to
   (easy to use)
 - you may provide a already opened socket at each call (performance)
 - WiSet provides __enter__ and __close__ methods, so it's easy to use
   in a pythonic way

However, some features are not yet implemented:
 - we only support IPv4 addresses
 - we don't support all ipset modules provided by recent kernels
 - only basic flags are currently supported

This code is running in production since a year on many devices. I'm
sorry to not provide a real git history of changes, but the project
began out of pyroute2 repository. The backport of code from our
externals libraries and pyroute2 project was not so easy.

Signed-off-by: Florent Fourcot <florent.fourcot@wifirst.fr>
Signed-off-by: Florian Vichot <florian.vichot@wifirst.fr>
Signed-off-by: Étienne Noss <etienne.noss@wifirst.fr>
Signed-off-by: Romain Bellan <romain.bellan@wifirst.fr>

### DIFF
--- a/pyroute2/wiset.py
+++ b/pyroute2/wiset.py
@@ -1,0 +1,480 @@
+'''
+WiSet module
+============
+
+High level ipset support.
+
+When :doc:`ipset` is providing a direct netlink socket with low level
+functions, a :class:`WiSet` object is built to map ipset objects from kernel. It
+helps to add/remove entries, list content, etc.
+
+For example, adding an entry with :class:`pyroute2.ipset.IPSet` object
+implies to set a various number of parameters:
+
+>>> ipset = IPSet()
+>>> ipset.add("foo", "1.2.3.4/24", etype="net")
+>>> ipset.close()
+
+When they are discovered by a :class:`WiSet`:
+
+>>> wiset = load_ipset("foo")
+>>> wiset.add("1.2.3.4/24")
+
+Listing entries is also easier using :class:`WiSet`, since it parses for you
+netlink messages:
+
+>>> wiset.content
+{'1.2.3.0/24': IPStats(packets=None, bytes=None, comment=None, timeout=None)}
+'''
+
+import errno
+import uuid
+from collections import namedtuple
+from inspect import getcallargs
+from socket import AF_INET
+
+from pyroute2 import IPSet
+from pyroute2.netlink.exceptions import IPSetError
+from pyroute2.netlink.nfnetlink.ipset import IPSET_FLAG_WITH_COUNTERS
+from pyroute2.netlink.nfnetlink.ipset import IPSET_FLAG_WITH_COMMENT
+
+
+# Debug variable to detect netlink socket leaks
+COUNT = {"count": 0}
+
+
+def need_ipset_socket(fun):
+    """ Decorator to create netlink socket if needed.
+
+    In many of our helpers, we need to open a netlink socket. This can
+    be expensive for someone using many times the functions: instead to have
+    only one socket and use several requests, we will open it again and again.
+
+    This helper allow our functions to be flexible: the caller can pass an
+    optional socket, or do nothing. In this last case, this decorator
+    will open a socket for the caller (and close it after call)
+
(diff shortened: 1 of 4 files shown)


pyroute2/wiset.py
```diff
@@ -0,0 +1,477 @@
...
+from pyroute2.netlink.nfnetlink.ipset import IPSET_FLAG_WITH_COUNTERS
+from pyroute2.netlink.nfnetlink.ipset import IPSET_FLAG_WITH_COMMENT
+
+
+# Debug variable to detect netlink socket leaks
+COUNT = {"count": 0}
+
+
+def need_ipset_socket(fun):
+    """ Decorator to create netlink socket if needed.
```
svinota: In the future it may be reasonable to move it to the common module and use not only for ipset.

pyroute2/wiset.py
```diff
@@ -0,0 +1,477 @@
...
+    >>>    myset.create()        # add the ipset in the kernel
+    >>>    myset.add("8.8.8.8")  # add this IP of google to the set
+
+    Netlink sockets are opened by __enter__ and __exit__ function, so you don't
+    have to manage it manually if you use the "with" keyword.
+
+    If you want to manage it manually (for example for long operation in
+    a daemon), you can do the following:
+
+    >>> myset = Wiset(name="mysuperipset")
```
svinota: `Wiset` or `WiSet`? It may be better to keep it consistent.

pyroute2/wiset.py
```diff
@@ -0,0 +1,477 @@
...
+                wiset.socket = socket
+            res[wiset.name] = wiset
+        elif content:
+            res[wiset.name].update_dict_content(myset)
+    return res
+
+
+@need_ipset_socket
+def list_ipset(name, content=False, socket=None, inherit_socket=False):
+    """ Get one ipset as Wiset object
```
svinota: A stupid comment — would it be possible to name it like `load_ipset()`? (just as a proposal to be discussed)

Another note: from my perspective it would be better not to name two different methods with only one letter difference in the names — too easy to make a mistake. I would propose `load_ipset()` and `load_all_ipsets()` or, at least, `list_ipset()` and `list_all_ipsets()`

pyroute2/wiset.py
ffourcot: Indeed, I will fix that

pyroute2/wiset.py
ffourcot: Yes, your are right. I will probably rename "socket" everywhere as well, since it can be confusing with the main "import socket" module.

svinota: +1 about socket also.

pyroute2/wiset.py
ffourcot: Yes, I will move it when merging the routing/ip part.

Otherwise I updated the pull request with other comment and I added documentation support (it adds a small conflict on index.rst, easy to solve)

svinota:
Grattis! The module is merged. Take care of it :)

PR #423, opened 2017-09-15
Ipset port fixes

AZaugg:
A few additional changes needed:

 - When setting hash:ports we must set a protocol with it, we must never specify a protocol when setting a bitmap:port
 - entry could be a string when setting an individual port. casting to int, in the event PortEntry is not used.

### DIFF
--- a/pyroute2/ipset.py
+++ b/pyroute2/ipset.py
@@ -243,6 +243,10 @@ class IPSet(NetlinkSocket):
 
     def _entry_to_data_attrs(self, entry, etype, ip_version):
         attrs = []
+
+        # Set the default protocol to tcp for hash:ports
+        p = lambda x: socket.getprotobyname('tcp') if x == None else x
+
         # We support string (for one element, and for users calling this
         # function like a command line), and tupple/list
         if isinstance(entry, basestring):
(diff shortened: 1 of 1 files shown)


pyroute2/ipset.py
```diff
@@ -243,6 +243,10 @@ def _family_to_version(family):
 
     def _entry_to_data_attrs(self, entry, etype, ip_version):
         attrs = []
+
+        # Set the default protocol to tcp for hash:ports
+        p = lambda x: socket.getprotobyname('tcp') if x == None else x
```
svinota: could it be solved simply by default values of `PortRange.protocol` and `PortEntry.protocol` ?

ffourcot: No, since several ipset modules don't support protocol options. And I'm not sure of support of all modules revisions of hash:ports, by the way.

Personally, I'm against setting default values here. They are too many modules/revisions combinations, we will probably break something. That should be done in upper layers, if someone need it.

svinota: Ok, reasonable.

I'm just thinking how to avoid lambda assignment here.

ffourcot:
@AZaugg : did you really need this feature or can we close this pull request? I think that it's a bad idea to merge it because:

- the cast to int probably does not hurt, but could be done by the caller. I don't have a strong opinion on it, but i don't see the benefit today
- your code to provide a default value for "hash:net,port and friends" looks like a hack. Your check on etype variable to provide a default value is working since only bitmap:port don't need today a protocol (and refuse if any is set), and it's the only type with port as single dimension. But if a new "foo:port" type is added in kernels, your code will be broken.

PR #425, opened 2017-09-23
Adds seg6 support

pierventre:
Changes:
- Introduces IPRoute support for seg6 tunnel
- Adds encap and inline modes
- Supports hmac functionality for SRH
- Initial parsing of human friendly form
- Add seg6_encap_info for RTA_ENCAP mgmt

### DIFF
--- a/pyroute2/netlink/rtnl/req.py
+++ b/pyroute2/netlink/rtnl/req.py
@@ -15,7 +15,8 @@ from pyroute2.netlink.rtnl.fibmsg import FR_ACT_NAMES
 
 
 encap_types = {'mpls': 1,
-               AF_MPLS: 1}
+               AF_MPLS: 1,
+               'seg6': 5}
 
 
 class IPRequest(dict):
@@ -119,6 +120,47 @@ class IPRouteRequest(IPRequest):
             if override_bos:
                 ret[-1]['bos'] = 1
             return {'attrs': [['MPLS_IPTUNNEL_DST', ret]]}
+        '''
+        Seg6 encap header transform. Format samples:
+
+            {'type': 'seg6',
+             'mode': 'encap',
+             'segs': '2000::5,2000::6'}
+
+            {'type': 'seg6',
+             'mode': 'encap'
+             'segs': '2000::5,2000::6',
+             'hmac': 1}
+        '''
+        if header['type'] == 'seg6':
+            # Init step
+            ret = {}
+            # Parse segs
+            segs = header['segs']
+            # If they are in the form in_addr6,in_addr6
+            if isinstance(segs, basestring):
+                # Create an array with the splitted values
+                temp = segs.split(',')
+                # Init segs
+                segs = []
+                # Iterate over the values
+                for seg in temp:
+                    # Discard empty string
+                    if seg != '':
+                        # Add seg to segs
+                        segs.append(seg)
+            # Retrieve mode
+            mode = header['mode']
+            # hmac is optional and contains the hmac key
+            hmac = header.get('hmac', None)
+            # Construct the new object
+            ret = {'mode': mode, 'segs': segs}
+            # If hmac is present convert to u32
+            if hmac:
+                # Add to ret the hmac key
+                ret['hmac'] = hmac & 0xffffffff
+            # Done return the object
+            return {'attrs': [['SEG6_IPTUNNEL_SRH', ret]]}
 
     def mpls_rta(self, value):
         ret = []
@@ -187,6 +229,29 @@ class IPRouteRequest(IPRequest):
                 #
                 # 'type' is mandatory
                 if 'type' in value and 'labels' in value:
+                    dict.__setitem__(self, 'encap_type',
+                                     encap_types.get(value['type'],
+                                                     value['type']))
+                    dict.__setitem__(self, 'encap',
+                                     self.encap_header(value))
+                # human-friendly form:
+                #
+                # 'encap': {'type': 'seg6',
+                #           'mode': 'encap'
+                #           'segs':'2000::5,2000::6'}
+                #
+                # 'encap': {'type': 'seg6',
+                #           'mode': 'encap'
+                #           'segs':'2000::5,2000::6'
+                #           'hmac': 1}
+                #
+                # 'encap': {'type': 'seg6',
+                #           'mode': 'encap'
+                #           'segs':'2000::5,2000::6'
+                #           'hmac': 0xf}
+                #
+                # 'type', 'mode' and 'segs' are mandatory
+                if 'type' in value and 'mode' in value and 'segs' in value:
                     dict.__setitem__(self, 'encap_type',
                                      encap_types.get(value['type'],
                                                      value['type']))
--- a/pyroute2/netlink/rtnl/rtmsg.py
+++ b/pyroute2/netlink/rtnl/rtmsg.py
@@ -160,7 +160,144 @@ class rtmsg_base(nlflags):
         __slots__ = ()
 
         nla_map = (('SEG6_IPTUNNEL_UNSPEC', 'none'),
-                   ('SEG6_IPTUNNEL_SRH', 'hex'))
+                   ('SEG6_IPTUNNEL_SRH', 'ipv6_sr_hdr'))
+
+        class ipv6_sr_hdr(nla):
+
+            __slots__ = ()
+
+            fields = (('encapmode', 'I'),
+                      ('nexthdr', 'B'),
+                      ('hdrlen', 'B'),
+                      ('type', 'B'),
+                      ('segments_left', 'B'),
+                      ('first_segment', 'B'),
+                      ('flags', 'B'),
+                      ('reserved', 'H'),
+                      ('segs', 's'),
+                      # Potentially several type-length-value
+                      ('tlvs', 's'))
+
+            # Corresponding values for seg6 encap modes
+            SEG6_IPTUN_MODE_INLINE = 0
+            SEG6_IPTUN_MODE_ENCAP = 1
+
+            # Mapping string to nla value
+            encapmodes = {
+              "inline": SEG6_IPTUN_MODE_INLINE,
+              "encap": SEG6_IPTUN_MODE_ENCAP
+            }
+
+            # Nla value for seg6 type
+            SEG6_TYPE = 4
+
+            # Flag value for hmac
+            SR6_FLAG1_HMAC = 1 << 3
+
+            # Tlv value for hmac
+            SR6_TLV_HMAC = 5
+
+            # Utility function to get the family from the msg
+            def get_family(self):
+                pointer = self
+                while pointer.parent is not None:
+                    pointer = pointer.parent
+                return pointer.get('family', AF_UNSPEC)
+
+            def encode(self):
+                # Retrieve the family
+                family = self.get_family()
+                # Seg6 can be applied only to IPv6
+                if family == AF_INET6:
+                    # Get mode
+                    mode = self['mode']
+                    # Get segs
+                    segs = self['segs']
+                    # Get hmac
+                    hmac = self.get('hmac', None)
+                    # With "inline" mode there is not
+                    # encap into an outer IPv6 header
+                    if mode == "inline":
+                        # Add :: to segs
+                        segs.insert(0, "::")
+                    # Add mode to value
+                    self['encapmode'] = self.encapmodes.get(mode, self.SEG6_IPTUN_MODE_ENCAP)
+                    # Calculate srlen
+                    srhlen = 8 + 16 * len(segs);
+                    # If we are using hmac we have a tlv as trailer data
+                    if hmac:
+                        # Since we can use sha1 or sha256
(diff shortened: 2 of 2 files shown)


pyroute2/netlink/rtnl/req.py
```diff
@@ -119,6 +120,27 @@ def encap_header(self, header):
             if override_bos:
                 ret[-1]['bos'] = 1
             return {'attrs': [['MPLS_IPTUNNEL_DST', ret]]}
+        '''
+        Seg6 encap header transform. Format samples:
+
+            {'type': 'seg6',
+            'mode': 'encap',
+            'segs':'2000::5,2000::6'}
```
svinota: how many segments there may be?

pyroute2/netlink/__init__.py
```diff
@@ -1678,6 +1678,79 @@ class mpls_target(target):
...
+                # Get mode
+                mode = encap.get('mode', None)
+                # Get segs
+                segs = encap.get('segs', [])
+                # Initialize value
+                self['value'] = b''
+                # Add mode to value
+                self['value'] += struct.pack('i', self.encapmodes.get(mode, 1))
+                # Nexthdr is not set
+                nexthdr = 0
```
svinota: is it always == 0?

pyroute2/netlink/__init__.py
```diff
@@ -1678,6 +1678,79 @@ class mpls_target(target):
...
+                # Calculate srlen
+                srhlen = 8 + 16 * len(segs);
+                # Calculate hdrlen
+                hdrlen = (srhlen >> 3) - 1;
+                # Add hdrlen to value
+                self['value'] += struct.pack('B', hdrlen)
+                # Add type to value
+                self['value'] += struct.pack('B', self.SEG6_TYPE)
+                # Calculate segments left
+                segments_left = len(segs) - 1;
```
svinota: what if we have 10 segments?

pyroute2/netlink/__init__.py
```diff
@@ -1678,6 +1678,79 @@ class mpls_target(target):
...
+                # Add segments left to value
+                self['value'] += struct.pack('B', segments_left)
+                # Calculate first segment
+                first_segment = len(segs) - 1
```
svinota: same question, what if we have 10 segments?

pyroute2/netlink/__init__.py
```diff
@@ -1678,6 +1678,79 @@ class mpls_target(target):
...
+                # Add segments first segment to value
+                self['value'] += struct.pack('B', first_segment)
+                # flags is not set for now
+                flags = 0
+                # Add flags to value
+                self['value'] += struct.pack('B', flags)
```
svinota: what are these flags, btw?

pyroute2/netlink/__init__.py
```diff
@@ -1678,6 +1678,79 @@ class mpls_target(target):
...
+                # reserved is not set for now
+                reserved = 0
+                # Add flags to value
+                self['value'] += struct.pack('H', reserved)
+                # Iterate over segments
+                for seg in segs:
```
svinota: so, the rest is simply concatenated packed IPv6 addresses?

pyroute2/netlink/rtnl/req.py
pierventre: AFAIK, the number can be very high. I think, looking at the elements of the header, it can be at most 256

pyroute2/netlink/__init__.py
pierventre: Long story short, they used the SR header (Segment Routing) of the IP packets to instruct the kernel about the tunnel. Nexthdr makes sense only for IP packets where after the IP header there is always something else. Instead for netlink it is not used.

pyroute2/netlink/__init__.py
pierventre: I didn't get. Why 10 should be a problem ? 

Btw, ";" is not necessary

pyroute2/netlink/__init__.py
pierventre: Same as above

pyroute2/netlink/__init__.py
pierventre: Flags are different from zero only if HMAC is used. They are in the todo-list at the beginning of the method.

pyroute2/netlink/__init__.py
pierventre: Yes, the segment routing header is just a list of waypoints (in jargon list of segments) we have to cross in the network. A segment is an IPv6 address.

pyroute2/netlink/__init__.py
svinota: Just trying to understand, to which segment does it point.

pyroute2/netlink/rtnl/req.py
```diff
@@ -119,6 +120,27 @@ def encap_header(self, header):
...
+        '''
+        if isinstance(header['type'], int) or \
+                (header['type'] in ('seg6')):
+            ret = []
```
svinota: Btw, why it is a list here?

pyroute2/netlink/__init__.py
pierventre: https://git.kernel.org/pub/scm/linux/kernel/git/shemminger/iproute2.git/tree/ip/iproute_lwtunnel.c#n467

pyroute2/netlink/rtnl/req.py
pierventre: I did not know very well the library so I just followed mpls code.

pyroute2/netlink/__init__.py
svinota: Ok, thanks!

pierventre:
Besides the missing functionalities (I will add asap) and the decode function that is missing, I would like to know what are the other parts I have to improve to finish the patch. Can you please point me where to start looking for ?

svinota:
I just thinking, what's better — to merge the code as is and to refactor it later using the standard library routines, or to discuss here with examples and merge the code thereafter.

pierventre:
Ok, please let me know. I am willing to help

svinota:
Actually, the NLA data could be decoded using the fields framework, as for any other NLA type. The only custom code would be then to turn the trailing string into a list of IPs and vice versa:
```python
162     class seg6_encap_info(nla):                                                  
163                                                                                  
164         __slots__ = ()                                                           
165                                                                                  
166         nla_map = (('SEG6_IPTUNNEL_UNSPEC', 'none'),                             
167                    ('SEG6_IPTUNNEL_SRH', 'ipv6_sr_hdr'))                         
168                                                                                  
169         class ipv6_sr_hdr(nla):                                                  
170             __slots__ = ()                                                       
171             fields = (('encapmode', 'I'),                                        
172                       ('nexthdr', 'B'),                                          
173                       ('hdrlen', 'B'),                                           
174                       ('type', 'B'),                                             
175                       ('segments_left', 'B'),                                    
176                       ('first_segment', 'B'),                                    
177                       ('flags', 'B'),                                            
178                       ('reserved', 'H'),                                         
179                       ('segs', 's'))
```

Now the `decode()`:
1. run `nla.decode()`
2. load the IP list from the binary data and place into segs

The `encode()`:
1. fill all the required values
2. put the encoded list as a string into `segs`
3. run `nla.encode()`

pierventre:
I have done what you suggested.

Should we remove the hardcoded "21" here?
https://github.com/svinota/pyroute2/blob/74db7d344f8dba8cd81f1713f77b5dafbb0fb579/pyroute2/netlink/rtnl/rtmsg.py#L139

What is the next step ?

svinota:
Yes, I'm to fix the hardcoded constant.

The code looks OK, then next steps could be:
1. the test routine
2. IPDB integration
3. test routines for the IPDB

I would merge the code now, and proceed with 1.

pierventre:
Ok, let me finish the functionalities for ip route and then we can move forward. In this way at least we have completed seg6 support for IPRouteRequest.

pierventre:
I have to add the support for inline mode, hmac and to properly fill the flags

svinota:
Ok, then I'll wait. Meanwhile I prepare the tests to commit asap.

svinota:

pyroute2/netlink/rtnl/rtmsg.py
```diff
@@ -160,7 +160,90 @@ class seg6_encap_info(nla):
...
+                # Decode the data
+                nla.decode(self)
+                # Calculate offset of the segs
+                offset = self.offset + 16
+                # Point the addresses
+                addresses = self.data[offset:]
+                # Init segs
+                segs = []
+                # Move 128 bit in each step
+                for i in range(len(addresses) // 16):
```
svinota: We should extract here `segments_left + 1`, since `self.data`  contains all the message, not only this NLA. Something like:
```python
                 for i in range(len(addresses) // 16):                            
                     if i > self['segments_left']:                                
                         break                                                    
                     # Save the segment                                           
                     segs.append(inet_ntop(AF_INET6, addresses[i*16:i*16+16]))    
```

pierventre: Yes, I was planning to do that in order to proper support HMAC functionality which introduces other data after segments.

https://git.kernel.org/pub/scm/linux/kernel/git/shemminger/iproute2.git/tree/ip/iproute_lwtunnel.c#n435

https://git.kernel.org/pub/scm/linux/kernel/git/shemminger/iproute2.git/tree/ip/iproute_lwtunnel.c#n455

pyroute2/netlink/rtnl/rtmsg.py
```diff
@@ -160,7 +160,144 @@ class seg6_encap_info(nla):
...
+                      ('nexthdr', 'B'),
+                      ('hdrlen', 'B'),
+                      ('type', 'B'),
+                      ('segments_left', 'B'),
+                      ('first_segment', 'B'),
+                      ('flags', 'B'),
+                      ('reserved', 'H'),
+                      ('segs', 's'),
+                      # Potentially several type-length-value
+                      ('tlvs', 's'))
```
pierventre: I am not sure if this is the better way to do that. What do you think ?

pyroute2/netlink/rtnl/rtmsg.py
```diff
@@ -160,7 +160,144 @@ class seg6_encap_info(nla):
...
+                    # Init segs
+                    self['segs'] = b''
+                    # Iterate over segments
+                    for seg in segs:
+                        # Convert to network byte order and add to value
+                        self['segs'] += inet_pton(family, seg)
+                    # Initialize tlvs
+                    self['tlvs'] = b''
+                    # If hmac is used we have to properly init tlvs
+                    if hmac:
```
pierventre: Maybe this part can be optimized. Any idea ?

pyroute2/netlink/rtnl/rtmsg.py
```diff
@@ -160,7 +160,144 @@ class seg6_encap_info(nla):
...
+                # Move 128 bit in each step
+                for i in range(n_segs):
+                    # Save the segment
+                    segs.append(inet_ntop(AF_INET6, addresses[i*16:i*16+16]))
+                # Save segs
+                self['segs'] = segs
+                # Init tlvs
+                self['tlvs'] = ''
+                # If hmac is used
+                if self.has_hmac():
```
pierventre: Same here

pyroute2/netlink/rtnl/rtmsg.py
svinota: Probably there is no better way right now, you're welcome to propose. This one also may not always work as expected, but I'll fix that.

pyroute2/netlink/rtnl/rtmsg.py
svinota: Any string concatenation via `+` is not optimal in Python, but we can optimize that later.

The most important is that the code is readable. It is much more important even than the performance, from my experience.

svinota:
I believe the code is ready to be merged — but it does not mean we should stop the work on optimizations etc. If you agree, I merge it and then provide the tests.

pierventre:
Ok sounds good to me. Let me know how I can help.

pyroute2/netlink/rtnl/req.py
```diff
@@ -192,6 +234,29 @@ def __setitem__(self, key, value):
                                                      value['type']))
                     dict.__setitem__(self, 'encap',
                                      self.encap_header(value))
+                # human-friendly form:
+                #
+                # 'encap': {'type': 'seg6',
+                #           'mode': 'encap'
+                #           'segs':'2000::5,2000::6'}
```
svinota: btw, it should work also as `'segs': ['2000::5', '2000::6']`, shouldn't it?

svinota:
Thanks!

pyroute2/netlink/rtnl/req.py
pierventre: yes, it works.

pierventre:
You're welcome. Let me know how I can help in future.

PR #442, opened 2017-12-07
Added support for list:set

AZaugg:


ffourcot:
Hello,

I'm merging it since it cannot hurt. However, I will try to do the following next week to really support this feature: 

- move  the "if" in the main loop. Even if "set" type is today only supported in one dimension, it's more logical
- add support of before/after options, probably with a new container class
- add tests

PR #443, opened 2017-12-11
Fix for IPRoute objects that can still be used after close in Python 2

etene:
Hello,

I noticed IPRoute contexts could still be used after being closed in Python 2. I assume I'm not the first one to stumble upon this, but it has caused inconsistent behavior between Python versions in our codebase so I figured it warranted a fix.

The gist of the issue is that `NetlinkSocket` stores references to the underlying socket methods, which can change (and do). So instead of storing references I used `__getattr__` to always return the "current" methods.

The reference-based approach caused IPRoute sockets to be in an inconsistent state when closed:
- the socket methods are replaced by dummies at closing time
- the stored reference still points to the old method
- the underlying os-level socket is not necessarily closed
Thus data still could be sent over the socket even if it was considered closed.

Only Python 2 is affected, because in Python 3 socket methods point to the "real thing" and are not replaced by dummies when closing.

I guess a `__getattr__` call could have a performance overhead compared to the old method, so tell me if this is a problem and I'll think of another way.

Cheers,
Étienne

### DIFF
--- a/pyroute2/netlink/nlsocket.py
+++ b/pyroute2/netlink/nlsocket.py
@@ -872,17 +872,6 @@ class NetlinkSocket(NetlinkMixin):
                                            SOCK_DGRAM,
                                            self.family,
                                            self._fileno)
-            for name in ('getsockname', 'getsockopt', 'makefile',
-                         'setsockopt', 'setblocking', 'settimeout',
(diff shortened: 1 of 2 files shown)


svinota:
Thanks! Testing

svinota:
Thanks!

Issue #444, opened 2017-12-15
IPRoute.get() can deadlock indefinitely due to internal exceptions


hetmankp:
Currently the method defined in netlink.nlsocket.NetlinkMixin.get() does not cleanly release locks when an exception is encountered internally (for example, an exception when calling recv() on the socket, etc). The result is that an IPRoute instance can enter a locked up state in which is it unable to receive any more messages via the get() method.

svinota:
Thanks, testing the PR

PR #445, opened 2017-12-15
Fix IPRoute.get() lockup due to internal exceptions

hetmankp:
The method defined in netlink.nlsocket.NetlinkMixin.get() does not
cleanly release locks when an exception is encountered internally (for
example, an exception when calling recv() on the socket, etc). The
result is that an IPRoute instance can enter a locked up state in which
it is unable to receive any more messages via the get() method.

This fixes issue #444

### DIFF
--- a/pyroute2/netlink/nlsocket.py
+++ b/pyroute2/netlink/nlsocket.py
@@ -598,171 +598,176 @@ class NetlinkMixin(object):
 
             ret = []
             enough = False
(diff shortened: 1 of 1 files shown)


svinota:
Yep. Thanks! Testing

svinota:
Looks like working w/o issues so far — merged/rebased manually. Thanks!

PR #453, opened 2018-02-07
Allow for size limits on list:set to be configurable

AZaugg:
Default limit on list:set is 8, this change will allow a user to set a different size limit on list:sets


ffourcot:
Thanks !

PR #457, opened 2018-02-11
Corrected syntax error for new attribute for create ipset

AZaugg:
Apologies, fixing my syntax error I introduced. I might look at writing some tests.


ffourcot:
@AZaugg we already some tests on ipset and wiset. I'm sorry, I forgot to run tests before the merge :(

PR #461, opened 2018-02-26
ipset: add skbinfo capability

atzm:
Hi,

This PR adds support for the `skbinfo` capability into the `ipset` which includes setting `skbmark`, `skbprio`, and `skbqueue` at adding entry.

Thanks.


svinota:
Thanks!

PR #468, opened 2018-03-14
Add setns syscall for s390

scheuran:
We need s390 support for creating network namespaces via pyroute2. Since OpenStack is exploiting pyroute2 [1], Neutron is broken on s390.

Would it be possible to trigger a release after this got merged? Thanks!

[1] https://github.com/openstack/neutron/commit/c4d4336e9ffc24b5d78691f0d45b3a5056c78330


scheuran:
Still checking that all tests succeed on s390...

scheuran:
There are a couple of issues in the gre and tun/tap area. My suggestion is to continue with this PR as the netns stuff is the most important. I will open another PR as soon as I got the gre and tuntap tests going...

scheuran:
Using this one instead: https://github.com/svinota/pyroute2/pull/469

Issue #470, opened 2018-03-23
wrong rtnl group flags constant names


rjarry:
Hi,

I have noticed that the rtnetlink multicast group flags (to use with `bind()` for kernels before 2.6.14) are declared with the names of the multicast group ids (to be used with `setsockopt(NETLINK_{ADD,DEL}_MEMBERSHIP)` since kernel 2.6.14). They should have a `RTMGRP_` prefix instead of `RTNLGRP_`.

https://github.com/svinota/pyroute2/blob/master/pyroute2/netlink/rtnl/__init__.py#L39
https://github.com/torvalds/linux/blob/master/include/uapi/linux/rtnetlink.h#L606

This is confusing and when trying to subscribe to a specific multicast group using the constants defined in `pyroute2.netlink.rtnl` with `setsockopt`, we get a `errno=22 invalid argument` error.

I know that renaming those constants would break the API but I feel this is necessary. I'll submit a PR with a draft patch for review.

svinota:
Yep, I already thought about it but didn't find it too serious — and looks like I may be wrong. 

Thanks, I'll wait for the PR and I believe we fix the issue in some way.

PR #471, opened 2018-03-23
rtnl: fix wrong rtmgrp constant names

rjarry:
The `RTNLGRP_` prefix is reserved for multicast group ids (to be used with `setsockopt()`). The group flags (legacy api to be used with bind()) are named with the `RTMGRP_` prefix.

Here is a link to the original linux commit that introduced the new api: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=ac6d439d2097

Rename the constants to use the correct prefix. And rename `RTNL_GROUPS` to `RTMGRP_DEFAULTS` for consistency.

Fixes: #470


svinota:
Actually, they aren't bad :) wrong — may be — but not that bad.

Thanks, I run tests to check all the things and we continue. Since 0.5.0 was cut yesterday (with some big IPDB changes), it may be a good idea to make such change now also.

pyroute2/netlink/rtnl/__init__.py
```diff
@@ -120,16 +120,16 @@
...
-    RTNLGRP_IPV6_IFADDR |\
-    RTNLGRP_IPV4_ROUTE |\
-    RTNLGRP_IPV6_ROUTE |\
-    RTNLGRP_IPV4_RULE |\
-    RTNLGRP_IPV6_RULE |\
-    RTNLGRP_NEIGH |\
-    RTNLGRP_LINK |\
-    RTNLGRP_TC |\
-    RTNLGRP_MPLS_ROUTE
+RTMGRP_DEFAULT_GROUPS = RTMGRP_IPV4_IFADDR |\
```
svinota: `RTMGRP_DEFAULTS`?

svinota:
Except of one typo I see no issues.

Give me some time to check dependencies in OpenStack also

svinota:
Oh no. Looks like there are more issues.

pyroute2/netlink/rtnl/__init__.py
rjarry: Ooops my bad :)

rjarry:
I did not manage to run the tests. I think I miss a lot of stuff in my environment, but I'm not sure what.

pyroute2/netlink/rtnl/__init__.py
```diff
@@ -120,16 +120,16 @@
...
-    RTNLGRP_IPV6_IFADDR |\
-    RTNLGRP_IPV4_ROUTE |\
-    RTNLGRP_IPV6_ROUTE |\
-    RTNLGRP_IPV4_RULE |\
-    RTNLGRP_IPV6_RULE |\
-    RTNLGRP_NEIGH |\
-    RTNLGRP_LINK |\
-    RTNLGRP_TC |\
-    RTNLGRP_MPLS_ROUTE
+RTMGRP_DEFAULTS = RTMGRP_IPV4_IFADDR |\
```
svinota: Found another issue. On the line 116 I map all the variables that start with `RTM` to create a dictionary.

It wasn't a problem with `RTNLGRP`, but with `RTMGRP` all the groups become mapped too. Could you please change on the line 116
```
map_namespace('RTM', globals())
```
to
```
map_namespace('RTM_', globals())
```
so it will not catch the groups?

thanks

svinota:
Pls look at one more comment

pyroute2/netlink/rtnl/__init__.py
rjarry: Oh sure, I missed that.

pyroute2/netlink/rtnl/__init__.py
```diff
@@ -113,23 +113,23 @@
 RTM_NEWSTATS = 92
 RTM_GETSTATS = 94
 RTM_NEWCACHEREPORT = 96
-(RTM_NAMES, RTM_VALUES) = map_namespace('RTM', globals())
+(RTM_NAMES, RTM_VALUES) = map_namespace('RTM_', globals())
```
svinota: Thanks!

svinota:
Checking the OpenStack components

svinota:
… OpenStack is ok. Merging

rjarry:
Thanks @svinota !

PR #475, opened 2018-04-03
(WIP) add nfconntrack count support

ffourcot:
Hello,

This patch is a proof of context (not to merge) of nfconntrack count command.

However, this request return only one netlink message, without any "DONE" message. So it's not detected by pyroute2 and hang forever.

I tried to add a "terminate" condition, but it does not work since it skips matching messages. I checked code of conntrack tools, and it uses special code to match this command and get the single message.

### DIFF
--- a/pyroute2/netlink/nfnetlink/conntrack.py
+++ b/pyroute2/netlink/nfnetlink/conntrack.py
@@ -36,6 +36,16 @@ IP_CT_TCP_FLAG_DATA_UNACKNOWLEDGED = 0x10
 # The field td_maxack has been set
 IP_CT_TCP_FLAG_MAXACK_SET = 0x20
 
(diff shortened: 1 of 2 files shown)


svinota:
Got it, will play with the code tonight.

Thanks!

svinota:
Please look at the `terminate` branch. If it helps, then we may go that way.

ffourcot:
It works for me, thanks. I updated and rebased the commit to use this new feature

svinota:
So I merge all into the master.

svinota:
Merged manually

svinota:

svinota:
The `NFCTSocket` class is re-exported from the root module.

From now on one can import it by
```
from pyroute2 import NFCTSocket
```

PR #476, opened 2018-04-08
Fix the file creation flag O_EXCL

tianquan23:
There's no flag named O_EXIST. It should be O_EXCL that ensures the call
fails if the namespace already exists.


svinota:
Indeed. Thanks!

PR #477, opened 2018-04-10
conntrack: add getter for cpu statistics

atzm:
Hi,

This PR adds getter for the cpu statistics into the conntrack.
It returns a list of dictionaries which contains the same values as command output of `conntrack -S`.

Thanks.

### DIFF
--- a/pyroute2/netlink/nfnetlink/conntrack.py
+++ b/pyroute2/netlink/nfnetlink/conntrack.py
@@ -48,6 +48,25 @@ class nfct_stats(nfgen_msg):
     )
 
 
+class nfct_stats_cpu(nfgen_msg):
+    nla_map = (
+        ('CTA_STATS_UNSPEC', 'none'),
+        ('CTA_STATS_SEARCHED', 'be32'),
+        ('CTA_STATS_FOUND', 'be32'),
+        ('CTA_STATS_NEW', 'be32'),
+        ('CTA_STATS_INVALID', 'be32'),
+        ('CTA_STATS_IGNORE', 'be32'),
+        ('CTA_STATS_DELETE', 'be32'),
+        ('CTA_STATS_DELETE_LIST', 'be32'),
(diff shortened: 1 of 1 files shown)


pyroute2/netlink/nfnetlink/conntrack.py
```diff
@@ -238,6 +257,17 @@ def dump(self, mark=None, mark_mask=0xffffffff):
         return self.request(msg, IPCTNL_MSG_CT_GET,
                             msg_flags=NLM_F_REQUEST | NLM_F_DUMP)
 
+    def stat(self):
```
svinota: Personally I would prefer to split the API in two parts:

1. pure netlink mapping, generic API that returns netlink messages as is w/o modifications
2. high-level utilities that may aggregate the info and transform it in some human-friendly manner

I believe it is important to export the raw netlink messages to allow people to make their own high-level APIs as well. So the netlink socket object would follow 1. and the utility class (that may be a child class to the previous one) would implement 2.

What would you say?

ffourcot: Hello,

Since count() is still not yet published in a pyroute2 release, it's fine for me as well to change it to expose raw netlink messages.

Do you prefer a new module (like WiSet for IPSet) for the high-level API?

svinota: @ffourcot  yep, I would add a new module under /pyroute2/, like wiset. Could you add it please?

@atzm, I merge this PR, but could you please also take a look at the new module when we add it, and probably provide the high level function there as well?

atzm: @svinota Sure.  Thank you for the work.

svinota:
@atzm @ffourcot please take a look at https://github.com/svinota/pyroute2/commit/56c8363e7b0860691021304d4a71b4db2c58f0d2

PR #478, opened 2018-04-11
nfnetlink: fix byte order of nfgen_msg.res_id field

atzm:
The convention in nfnetlink is to use network byte order in every header field
except for old nft implementation on the ancient kernel.  Also recent kernel
provided a workaround for old nft, so now en/decoding res_id field as network
byte order seems safe:

  https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=a9de9777d613500b089a7416f936bf3ae5f070d2


svinota:
Tested with the `NFTables` module, looks like working. Thanks!

PR #482, opened 2018-04-14
nfctsocket: add support for entry deletion

atzm:
Hi,

This PR adds support for conntrack entry deletion.  It provides pure mapping for kernel API.
If the delete function by "search and destroy" like conntrack command is needed, I believe conntrack module should provide it.

Thanks.

### DIFF
--- a/pyroute2/netlink/nfnetlink/nfctsocket.py
+++ b/pyroute2/netlink/nfnetlink/nfctsocket.py
@@ -8,6 +8,7 @@ import socket
 
 from pyroute2.netlink import NLM_F_REQUEST
 from pyroute2.netlink import NLM_F_DUMP
+from pyroute2.netlink import NLM_F_ACK
 from pyroute2.netlink import NETLINK_NETFILTER
 from pyroute2.netlink import nla
 from pyroute2.netlink.nlsocket import NetlinkSocket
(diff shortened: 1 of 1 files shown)


atzm:
Agreed with keeping the manner as other modules.

IMHO it may be better to separate creating and sending request, to use common method to create request from `flush`, `dump`, and other instructions (probably `stat` and `count` can use it too).

atzm:
Sorry for the delay.

I tried to merge the approaches.  Could you please review it?

svinota:
Looks interesting, so I'm to merge it.

There are two things I would like to discuss before the merge.

1. I'm not sure that `count()` and `stat()` require an additional method call (here: +frame) as they run with an empty nfct message. Are there reasons to call `_mkmsg()` from those methods? (TBH, I'm not sure that the separate `_mkmsg()` is required at all, but I would rather trust you here)

2. Personally I do not like custom types like `NFCTAttrTuple`, but the type may be a good idea as it enforces the attrs, unlike a generic dict structure. Could you please elaborate on pros/contras for that? So far I see:
+ (+) Enforces attributes, strict type checking
+ (+) Makes the unit testing easier
+ (-) Requires an additional import in those modules that use the socket

atzm:
Thank you for the review.

1. Right.  I'll revert `count()` and `stat()`.

    Following the review, I feel it may be better to move `_mkmsg()` into the class `nfct_msg` as a class method named `create()`, `instance()`, `fromattrs()` or etc.  What do you think?

2. Yes.  Additionally:
    - (+) Improves readability and extendability.  Because probably how to create attribute structure from user data will differ depends on each CTA (nested nla in particular).

    Also this custom type can be merged into the each internal class `cta_*`, but then users must know about these internal classes and use them too, for example: `ct.entry('set', tuple_orig=nfct_msg.cta_tuple.attrs(saddr="192.168.2.1", ...))`
    Which would you like better?

svinota:
Yes, I believe it would be a good idea to move `_mkmsg()` into the `nfct_msg` class. Two comments:
1. Naming the method, please make sure there is no name overlap with the `nlmsg_base` class methods. Something like `create()` or `create_from()` should work. 
2. Later we could call the method from the constructor as well — another benefit

About the `cta_tuple`: I would say that the high level API should not require any `nfct_msg` import at all, so I would rather not merge the type into the message class, but leave as is at the module level, it should work as well.

Thanks for the contribution!

atzm:
Agreed, fixed with your comments.  Thank you for the review!

svinota:
Merged. I'll write some basic tests for that asap.

PR #499, opened 2018-05-08
Introduce "eventloop" style API to better handle netlink event storms

crosser:
Please consider this proposal.

In mass network topology updates, thousands of netlink packets lead to excessive resource usage (one thread per event use much memory and results in slow processing) and hanging process (recovery from failed IPDB commit is slow, and monitoring thread is prone to die without the master thread noticing).

This change allows the user to process incoming netlink messages in an event loop (in the main thread, or in any user-initiated thread(s)), and netlink socket receive errors are passed via the same queue, and re-raised in the master thread. This allows the user program to notice that netlink events where lost and take appropriate action.

The last commit of three allows to specify the size of netlink socket send and receive buffers, and thus control the "storm sensitivity" of the system.

Sidenote: on Linux, system wide max size of socket receive and send buffers is controlled with
/proc/sys/net/core/rmem_max, it may be worth to mention in the documentation.

### DIFF
--- a/pyroute2/ipdb/main.py
+++ b/pyroute2/ipdb/main.py
@@ -704,6 +704,10 @@ import logging
 import traceback
 import threading
 import weakref
+try:
+    import queue
+except ImportError:
+    import Queue as queue  # The module is called 'Queue' in Python2
 
 from functools import partial
 from pprint import pprint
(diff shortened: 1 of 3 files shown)


svinota:
@celebdor Antoni, pls take a look at the PR

PR #502, opened 2018-05-14
Trivial style fix, indentation was wrong

crosser:
Indentation of a block is wrong. It works correctly, but makes
the code difficult to read.


svinota:
Nope, this `else` corresponds to the `for` statement, not `if`.

The reason: sometimes the kernel does not send the updates properly, for that case I try to set up addresses several times (here 3). It does matter if you set some hundreds addresses on an interface.

So I'm to add a proper comment to that `else`.

crosser:
ok, ok. I was confused, sorry.

Issue #501, opened 2018-05-14
pyroute2.NetNS leaks FDs on non-existing Network Namespace


edwardbadboy:
Hi, recently I found a problem in the instantiating code of `pyroute2.NetNS`.

It's like
```
class NetNS(IPRouteMixin, RemoteSocket):
    def __init__(self, netns, flags=os.O_CREAT):
        self.netns = netns
        self.flags = flags
        self.cmdch, self._cmdch = config.MpPipe()
        self.brdch, self._brdch = config.MpPipe()
        atexit.register(self.close)
        self.server = config.MpProcess(target=NetNServer,
                                       args=(self.netns,
                                             self._cmdch,
                                             self._brdch,
                                             self.flags))
        self.server.start()
        super(NetNS, self).__init__()
        self.marshal = MarshalRtnl()
```

If the given Network Namespaces does not exist, then an OSError would be raised from the line `super(NetNS, self).__init__()`, for example:

```
>>> import pyroute2
>>> with pyroute2.NetNS('aaa') as ns:
...     print ns
...
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python2.7/dist-packages/pyroute2/netns/nslink.py", line 193, in __init__
    super(NetNS, self).__init__()
  File "/usr/local/lib/python2.7/dist-packages/pyroute2/netlink/nlsocket.py", line 296, in __init__
    super(NetlinkMixin, self).__init__()
  File "/usr/local/lib/python2.7/dist-packages/pyroute2/remote/__init__.py", line 253, in __init__
    raise init['error']
OSError: [Errno 1] mount rundir failed: 'aaa'
```

Then `self.cmdch` and self.`brdch` would remain open in this case, so FDs are leaked.

`self.cmdch` and `self.brdch` should be deleted and the FD should be closed automatically when the NetNS object is destroyed. However by using  `atexit.register(self.close)`, each instantiating of the class would keep a bound method in `atexit._exithandlers`, so there are reference to `self` of this bound method and `self.cmdch` and `self.brdch` are kept by `self`.

Even if a successful invocation of NetNS instantiation with a successful `close()` will leave a bounded method registered in atexit. I think it's kind of memory leak:
```
>>> import pyroute2
>>> import atexit
>>> with pyroute2.NetNS('nsx-node-agent') as ns: print ns
...
<pyroute2.netns.nslink.NetNS object at 0x7f1c1a9d2950>
>>> atexit._exithandlers
[...]

PR #503, opened 2018-05-16
Fix NetNS Resource Leakage

edwardbadboy:
Hi, would you take a look at the patch to fix issue 501? If it's OK I can add some unit test cases.

In this patch I still keep the atexit calls, but have the close function remove itself from the atexit registered method.

I think it's better to abandon the atexit dynamic registration, however had not come up a perfect solution for this yet.

Bug-Url: https://github.com/svinota/pyroute2/issues/501

### DIFF
--- a/pyroute2/netns/nslink.py
+++ b/pyroute2/netns/nslink.py
@@ -66,10 +66,8 @@ run `remove()`.
 
 import os
 import errno
-import fcntl
 import atexit
 import signal
-import sys
 import logging
 from pyroute2 import config
 from pyroute2.netlink.rtnl.iprsocket import MarshalRtnl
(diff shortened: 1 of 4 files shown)


pyroute2/netns/nslink.py
```diff
@@ -82,7 +82,7 @@
 log = logging.getLogger(__name__)
 
 
-def NetNServer(netns, cmdch, brdch, flags=os.O_CREAT):
+def NetNServer(netns, parent_cmdch, cmdch, parent_brdch, brdch, flags=os.O_CREAT):
```
svinota: Only a cosmetic issue: pls use flake8 to check the PEP8 compliance.

pyroute2/netns/nslink.py
```diff
@@ -181,22 +183,40 @@ class NetNS(IPRouteMixin, RemoteSocket):
...
+        except Exception:
+            self.close()
+            raise
+        atexit.register(self.close)
         self.marshal = MarshalRtnl()
 
     def clone(self):
         return type(self)(self.netns, self.flags)
 
+    def _cleanup_atexit(self):
```
svinota: Thanks for the hint, I have to fix that in IPDB as well.

svinota:
Looks like working. If you can add tests to `/tests/unit/` or `/tests/general/` — would be awesome. I use `/tests/general/test_stress.py` to test respawns and like that.

edwardbadboy:
Got it, thanks. I'll update the patch.

Issue #400, opened 2017-07-28
NetNS objects leak


nickhuber:
the NetNS object doesn't clean up all of its resources on close() leading to some memory and fd leaks.

This is pretty easily replicated:

```
import pyroute2
for _ in range(1000):
    n = pyroute2.NetNS('testing')
    n.close()
```

On my system if I run that again I'll get an error about too many open files

lsof shows many open pipes:

```
...
python3 7044 root    8r  FIFO               0,11       0t0 286015306 pipe
python3 7044 root    9r  FIFO               0,11       0t0 286085453 pipe
python3 7044 root   10r  FIFO               0,11       0t0 286085459 pipe
python3 7044 root   11r  FIFO               0,11       0t0 286085465 pipe
python3 7044 root   12r  FIFO               0,11       0t0 286084620 pipe
python3 7044 root   13r  FIFO               0,11       0t0 286089223 pipe
python3 7044 root   14r  FIFO               0,11       0t0 286085471 pipe
python3 7044 root   15r  FIFO               0,11       0t0 286087185 pipe
python3 7044 root   16r  FIFO               0,11       0t0 286087191 pipe
python3 7044 root   17r  FIFO               0,11       0t0 286086262 pipe
...
```

also if you monitor the memory usage of the interpreter it grows with each created NetNS object

svinota:
A bit strange, as it passes the test: https://github.com/svinota/pyroute2/blob/master/tests/general/test_stress.py#L87

It may mean also, that the test doesn't catch the bug — that's evern worse.

Thanks a lot for the feedback. Could you please specify also the library version you use?

nickhuber:
I'm using 0.4.18 of pyroute2 from pypi

svinota:
That's bad.

Thanks again! Investigating.

svinota:
Can not confirm the pipes leakage with the code provided neither with Python3, nor with Python2 — checked with `/proc/<pid>/fd/`. Both on 0.4.18 and on the master branch.

Could you pls confirm that you can observe leaked pipes in `/proc/<pid>/fd/` with the code provided? And what the Python3 version is used?

Probably, the leakage was caused by some other condition, so I continue the investigation.

With the memory it's a bit more complicated issue, I will report here a bit later.

nickhuber:
```
ls -al /proc/12490/fd
total 0
dr-x------ 2 root root  0 Jul 28 13:07 .
[...]

PR #504, opened 2018-05-18
Fix NetNS Resource Leakage

edwardbadboy:
Hi, I added unit tests for the patch, and also found that we don't need the workaround for  https://github.com/svinota/pyroute2/issues/400 anymore, so deleted related code in the patch.

Bug-Url: #501

### DIFF
--- a/pyroute2/netns/nslink.py
+++ b/pyroute2/netns/nslink.py
@@ -66,10 +66,8 @@ run `remove()`.
 
 import os
 import errno
-import fcntl
 import atexit
 import signal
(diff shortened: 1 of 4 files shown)


edwardbadboy:
ping ?

svinota:
Sorry for delay. Tested, excellent. Thanks for the contribution!

srijnair-cisco:
Which rpm should I get for centos7 and python 2.7  if I require this patch and the patch for ipdb leak? patch #455 and #504

PR #508, opened 2018-05-21
Etene py2 close fix

valerie1706:


svinota:
Ummm… What's that?

etene:
It has my username in it but I don't understand either...

PR #510, opened 2018-06-04
Allow to perform active scanning.

plorinquer:
The `ssids` parameter has been added to the `scan()` method. This allows to perform active scanning when a list of SSIDs is provided.


svinota:
Thanks!

PR #511, opened 2018-06-04
Allow to flush the cache before scanning.

plorinquer:
Add a `flush_cache` parameter to the `scan()` method, in order to flush the cache before scanning.


svinota:
Excellent. Merged

PR #519, opened 2018-06-22
Update README.md

gordio:
Add syntax highlight.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,21 +40,25 @@ extended API. The additional functionality aims to:
 
 Maybe the simplest usecase is to monitor events. Disk quota events::
 
-    from pyroute2 import DQuotSocket
-    # DQuotSocket automatically performs discovery and binding,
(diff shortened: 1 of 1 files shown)


svinota:
The issue is that the file is used not only as `.md`, but as `.rst` source as well to generate a web page: https://docs.pyroute2.org/general.html , where the syntax is highlighted automatically with the sphinx plugin. If I merge the patch as is, it will look nice on the github page, but on the docs page it will be like on the screenshot.

One more thing we have to care about is the pypi page that's also using the README: https://pypi.org/project/pyroute2/ 

I'm not sure how to solve that not breaking anything in the puzzle :) Have you any ideas — you're welcome, I'm not closing the PR yet.

svinota:
screenshot

gordio:
Oh. I understand and agree. And don't have solution :-(

PR #530, opened 2018-07-23
tc: fix size of packets

ffourcot:
/**
 * struct gnet_stats_basic - byte/packet throughput statistics
 * @bytes: number of seen bytes
 * @packets: number of seen packets
 */
struct gnet_stats_basic {
    __u64   bytes;
    __u32   packets;
};


svinota:
Thanks!

PR #534, opened 2018-08-13
rtmsg: Add mapping for ttl attribute in tunnelled mpls label

sworleys:
Tunnelled/encapped mpls labels can also have a TTL attribute
passed with them. This was getting marked as unknown since it
didn't know how to parse it. Added the proper mapping.


svinota:
Thanks!

PR #538, opened 2018-08-29
Don't assume libc version

hetmankp:
While glibc is currently at version 6, there are other libc projects in
use which implement an older interface version, such as uClibc. Instead
of explicitly specifying the libc name, search for any libc that is
available.


svinota:
Reasonable. Thanks

PR #544, opened 2018-09-25
ss2: extend for usr context association

cherusk:
Associating pertaining sockets with context holding those via fds.

Signed-off-by: matthias <matthias.tafelmeier@gmx.net>

### DIFF
--- a/cli/ss2
+++ b/cli/ss2
@@ -16,9 +16,12 @@
 # along with this program. If not, see <http://www.gnu.org/licenses/>.
 
 import json
+import re
+import os
 import argparse
 from socket import AF_INET
-from collections import Callable
+import collections
+import psutil
 from pyroute2 import DiagSocket
 from pyroute2.netlink.diag import (SS_ESTABLISHED,
                                    SS_SYN_SENT,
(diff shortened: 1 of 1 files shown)


cherusk:
Known violations:
cli/ss2:65:9: E731 do not assign a lambda expression, use a def
cli/ss2:68:9: E731 do not assign a lambda expression, use a def

Asking for tolerating those.

cli/ss2
```diff
@@ -35,12 +38,82 @@ from pyroute2.netlink.diag import (SS_ESTABLISHED,
...
+            self._data[sk_inode][usr][pid]["fds"].append(sk_fd)
+
+        if sk_inode in self._data.keys():
+            if usr in self._data[sk_inode].keys():
+                if pid not in self._data[sk_inode][usr].keys():
+                    _enter_pid(ctxt)
+                _enter_fd(sk_fd)
+            else:
+                self._data[sk_inode][usr] = {}
+                _enter_pid(ctxt)
```
svinota: Nevertheless, you use it here, and it's not only confusing, but also causes unnecessary variable transitions. Simply to ignore the argument in the function and use a variable with the same name from the upper context.

cli/ss2
```diff
@@ -35,12 +38,82 @@ from pyroute2.netlink.diag import (SS_ESTABLISHED,
...
+                _enter_fd(sk_fd)
+        else:
+            self._data[sk_inode] = {}
+            self._data[sk_inode][usr] = {}
+            _enter_pid(ctxt)
+            _enter_fd(sk_fd)
```
svinota: So if you believe the lambda usage is needed here (I don't), think about that:
1. If you need to call a lambda several times, one of the possible ways is to store lambdas in a dictionary. Technically it's the same as to define a class with methods. From the bytecode perspective it's almost the same.
2. Every variable used within the lambda should be passed as an argument, no implicit upper context lookups — simply 'cause it works a bit weird in some situations.


Here I would define two additional functions in some context, probably in the global context.

svinota:
I would not tolerate. Please, use ordinary function definitions, as here the lambda usage does not simply pays off.

svinota:
The function definitions may be in any scope, I believe, there is no press about speed optimizations. So it's totally up to you to define them within the method or in some upper scope.

cherusk:
In the end, you're right.

As to the performance, actually, even in this ss2 streams, one person once commented actually that aspect as to closures, getting reallocated, yet, holds true for lambdas all the same:

https://github.com/svinota/pyroute2/pull/485

Will unroll it, thanks so far.

svinota:
Thanks!

cherusk:
Resorted to recursion for the entering. Would even describe it as close to aesthetic by now. Advantages should be obvious.

Foreseeing a performance boost only from parallelizing bottle-neck code paths anyways, but that's preliminary.

svinota:
Looks ok. Thanks a lot!

svinota:
please notice that the test is failing: 
```
======================================================================
ERROR: test_cli.TestTools.test_ss2
----------------------------------------------------------------------
  File "/usr/lib/python3/dist-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/peet/Projects/pyroute2/tests/general/test_cli.py", line 283, in test_ss2
    tcp_flows_hive[what] = future.result()
  File "/usr/lib/python3.6/concurrent/futures/_base.py", line 425, in result
    return self.__get_result()
  File "/usr/lib/python3.6/concurrent/futures/_base.py", line 384, in __get_result
    raise self._exception
  File "/usr/lib/python3.6/concurrent/futures/thread.py", line 56, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/home/peet/Projects/pyroute2/tests/general/test_cli.py", line 260, in do_ss2
    self.ss2.run(args)
  File "./bin/ss2", line 465, in run
    if args.process:
AttributeError: type object 'args' has no attribute 'process'
```

svinota:
nevermid, fixed.

PR #545, opened 2018-10-05
ss2: introduce name resolving

cherusk:
Further plain functionality.


svinota:
Here it was trivial, so I just fixed it: https://github.com/svinota/pyroute2/commit/88eee3e09747eb48d7c439a3af0e216d138c95ec

PR #560, opened 2018-11-02
Add retries of failed nlm_request calls for error code 16

tomkcook:
This does up to three retries of nlm_request when it fails with error code 16 ("Device or resource busy"), sleeping for 300ms between attempts.

This error seems to happen when multiple requests happen in quick succession, so slowing things down slightly when it happens seems to work.  I haven't spent any time experimenting with shorter sleeps - shorter times may also work or may work less reliably.

### DIFF
--- a/pyroute2/netlink/nlsocket.py
+++ b/pyroute2/netlink/nlsocket.py
@@ -829,29 +829,39 @@ class NetlinkMixin(object):
 
         msg_seq = self.addr_pool.alloc()
         with self.lock[msg_seq]:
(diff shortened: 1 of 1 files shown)


tomkcook:
I do still very occasionally see failures with this patch.  I'm reluctant to make the sleep a lot longer, or to make the number of retries infinite - is there a case where this error could be returned permanently?

svinota:
A good question. Let's do not make it infinite, but up to 30 retries looks ok.

PR #565, opened 2018-11-22
Fix bug in seg6 encap decode(), add seg6local support

cscarpitta:
Changes:
- Fix bug in seg6 decode(): flush_routes() give an error when it tried to delete all the routes because encapmode field was not decoded
- Introduces IPRoute support for seg6local tunnel
- Adds behaviours: End, End.X, End.T, End.DX2, End.DX6, End.DX4, End.DT6, End.B6, End.B6.Encaps
- Supports hmac functionality for SRH
- Initial parsing of human friendly form

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,6 +148,43 @@ Some examples::
                     'segs':'2000::5,2000::6,2000::7,2000::8',
                     'hmac':0xf})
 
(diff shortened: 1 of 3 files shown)


svinota:
Merged manually. Pls use `flake8` to check the code for the PEP8 compliance. Thanks!

PR #569, opened 2018-11-29
ipset: add GET_BYNAME and GET_BYINDEX commands

Vic063:


ffourcot:
Merged, this feature will be available in kernel >= 5.0: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/net/netfilter/ipset/ip_set_core.c?h=v5.0-rc3#n1868

PR #571, opened 2018-12-12
Replace "iteritems" with "items"

ralonsoh:
Replace Python2 dict.iteritems() with Py2/Py3 dict.items(),
to avoid Python3 errors.


celebdor:
Thanks for the change Rodolfo

PR #572, opened 2018-12-16
rtnl: Add Support for XFRM Interfaces

seglberg:
Virtual XFRM interfaces are available in Kernel 4.19+

Signed-off-by: Matt Ellison <matt@arroyo.io>

---
This change is [Reviewable](https://reviewable.io/reviews/svinota/pyroute2/572)

### DIFF
--- a/pyroute2/ipdb/main.py
+++ b/pyroute2/ipdb/main.py
@@ -1317,7 +1317,7 @@ class IPDB(object):
             # intefaces
             kind = target.get('kind', None)
             if kind in ('vlan', 'vxlan', 'gre', 'tuntap', 'vti', 'vti6',
-                        'vrf'):
+                        'vrf', 'xfrm'):
                 tx_prio1.append((target, tx))
             elif kind in ('bridge', 'bond'):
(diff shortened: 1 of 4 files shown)


celebdor:
@seglberg Does this PR aim to support also policy operations or just creation?

seglberg:
This is actually for a newer type of interface (also called xfrm) which can be used to send/receive traffic directly from xfrm policies using a lookup key (IF_ID).

This doesn't aim to add support for xfrm policies itself since in our use case StrongSwan is creating those for us. We just need to create the corresponding interfaces.

svinota:
The PR is missing the definition file (I believe under `ifinfmsg/plugins/`) that describes `IFLA_XFRM_…`

You can provide it here, or let me know so I add it myself, it's 5 minutes work.

Sorry for the delay, I couldn't work with any code quite some time.

seglberg:
Looks like I had them but forgot to commit... strange. I'll get them pushed.

svinota:
Thanks! Looks like it doesn't break anything, I'm merging.

Issue #578, opened 2019-01-19
NetNS file descriptors are closed before stopping any other thread using them


ralonsoh:
(Related to https://github.com/svinota/pyroute2/issues/547)

In OpenStack Neutron, pyroute2 is used inside a privsep context, to be able to execute root actions. Last oslo.privsep version (1.31.0), introduced the following feature [1]. We found several errors when executing pyroute2 commands when NetNS class is involved [2]. Occasionally, the following error is thrown [3].

We solved this issue by:
- Stopping gracefully the NetNS server instead of killing it.
- Stopping the recv and read loops before closing the file descriptor.

The following patch is being tested in the OpenStack CI successfully [4].

[1] https://github.com/openstack/oslo.privsep/commit/f368430f13183c3d09d68c80184d854449da2e87
[2] https://bugs.launchpad.net/neutron/+bug/1811515
[3] http://logs.openstack.org/92/628492/2/check/neutron-tempest-plugin-dvr-multinode-scenario/8c51f22/controller/logs/screen-q-l3.txt.gz#_Jan_13_18_22_39_027287
[4] https://github.com/ralonsoh/pyroute2/commit/e475d442bec8cb17f8868ec7a384b4940cd64da3

ralonsoh:
CI job for testing the patch [4] related in the previous comment: https://review.openstack.org/#/c/631075/

svinota:
Got it. Testing

svinota:
Looks like it doesn't break anything. Would you mind submitting a PR?

PR #579, opened 2019-01-22
Stop NetNS server gracefully

ralonsoh:
This patch changes the way of stopping the NetNS server, by sending
a SIGTERM signal and waiting for the server loop to finish. When the
signal is received, the loop control flag is inverted. Once the loop
is finished, the server process ends.

This patch also modifies the Transport class receiver function loops.
When the Transport object is closed, the receiver loops are stopped but
not the file descriptors. Those file descriptors, created in the NetNS
parent class, are closed at the end of the NetNS.close function, once
the child process (Server) is finished and the Transport receiver loops
are stopped. At this point, the file descriptors are not in use and can
be closed.

### DIFF
--- a/pyroute2/netns/nslink.py
+++ b/pyroute2/netns/nslink.py
@@ -167,6 +167,8 @@ class NetNS(RTNL_API, RemoteSocket):
                 os._exit(0)
 
         try:
+            self.remote_trnsp_in.close()
+            self.remote_trnsp_out.close()
             super(NetNS, self).__init__(trnsp_in, trnsp_out)
         except Exception:
             self.close()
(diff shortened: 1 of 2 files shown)


ralonsoh:
Related issue: https://github.com/svinota/pyroute2/issues/578

celebdor:
@ralonsoh is it only related or is it supposed to close it? If the latter, please, modify the commit message so it has:
```
closes #578
```

ralonsoh:
> @ralonsoh is it only related or is it supposed to close it? If the latter, please, modify the commit message so it has:
> 
> ```
> closes #578
> ```

Done

edwardbadboy:
Seems the check failure is not related to the patch. I can see error messages like:
```
./tests/run.sh: line 33: cd: /home/travis/build/svinota/pyroute2/dist: No such file or directory
tar: benchmark: Cannot read: Is a directory
tar: At beginning of tape, quitting now
tar: Error is not recoverable: exiting now
mv: cannot stat ‘pyroute2*/pyroute2’: No such file or directory
```
in https://travis-ci.org/svinota/pyroute2/jobs/482945874 .

ralonsoh:
> I see that in pyroute2/iproute/remote.py, the class RemoteIPRoute is also using Transport. If Transport doesn't close the FDs, maybe there will be resource leak in RemoteIPRoute ?

svinota:
Looks like working, I merge it.

PR #583, opened 2019-01-22
iproute: add support of NOARP flag

ffourcot:
NOARP is only a flag, that one can set with sysfs (with
/sys/class/net/<dev>/flags file) or using interface flag with netlink.

This commit adds a helper to mimic ip commands:
 * ip link set arp on dev <dev>
 * ip link set arp off dev <dev>

It's now possible to do:
 ip.link('set', arp=True, index=ifindex)
 ip.link('set', arp=False, index=ifindex)


svinota:
Thanks!

PR #586, opened 2019-02-19
Clear packets in raw socket buffer after bind()'ing to an interface or prior to attaching a filter

andrewstrohman:
Otherwise, you can recv packets that you don't expect.

### DIFF
--- a/pyroute2/protocols/rawsocket.py
+++ b/pyroute2/protocols/rawsocket.py
@@ -68,12 +68,13 @@ class RawSocket(socket):
             self.clear_buffer(remove_total_filter=True)
 
     def clear_buffer(self, remove_total_filter=False):
-        #there is a window of time after the socket has been created and before
-        #bind/attaching filter where packets can be queued onto the socket buffer
-        #see comments in function set_kernel_filter() in libpcap's pcap-linux.c
-        #libpcap sets a total filter which does not match any packet.
-        #It then clears what is already in the socket
-        #before setting the desired filter
+        # there is a window of time after the socket has been created and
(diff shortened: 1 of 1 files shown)


svinota:
Reasonable.

Could you please fix the code to be PEP8-compliant? Just use the latest flake8, there are some minor issues.

Otherwise I can do it self tomorrow.

Thanks!

svinota:
Cool. Thanks!

andrewstrohman:
@svinota, I just noticed an unexpected side-effect of this change.    If the network interface is down at RawSocket() creation time, the recvfrom() call that is used to clear the socket buffer raises an exception(ENETDOWN).   

This unfortunate, as the user may not expect the to receive that type of exception when creating a socket(they may expect it to only to happen on send/recv calls).   More specifically,  I would like to be able to create this RawSocket regardless if the interface is up or not.

I seem to only receive ENETDOWN the first time I call recvfrom on a down interface.   Thereafter, recvfrom just blocks until the interface comes back up and a packet arrives.

I'm tempted to catch and discard the ENETDOWN exception in the clear_buffer recvfrom loop, but this approach may be problematic for users who want receive this exception when they call recv for the first time.

Do you have any suggestions on how to handle this?

svinota:
I agree. Personally I would prefer not to raise any not critical exception from the socket object constructor.  In other words, if a socket object may be created, it should be created.

If that means we need to ignore ENETDOWN in `clear_buffer()`, I would probably do that way.

But I would like to see that as a note in the docstring, documenting this behaviour. Thank you!

PR #587, opened 2019-02-20
Syntax fix: use correct syntax to set default protocol to ETH_P_ALL i…

Vic063:
…n TC classifiers

Same as #585


svinota:
Tests passed. Thanks!

PR #588, opened 2019-02-27
fix parsing of options without policy

andrewstrohman:
Decoding option 61 was triggering exceptions due to incorrect parsing.


svinota:
Oh sure. Thank you a lot for contributing!

PR #590, opened 2019-03-01
netns: add documentation on how to go back on default netns

ffourcot:
Hello,

I didn't find any reference in the documentation for this annoying thing. There is perhaps a better way, but in all cases I think that it's a good idea to add documentation on it.

### DIFF
--- a/pyroute2/netns/__init__.py
+++ b/pyroute2/netns/__init__.py
@@ -222,6 +222,11 @@ def setns(netns, flags=os.O_CREAT, libc=None):
         - O_CREAT -- create netns, if doesn't exist
         - O_CREAT | O_EXCL -- create only if doesn't exist
 
+    Note that "main" netns as no name. But you can access it with::
+
+        setns('foo')  # move to netns foo
+        setns('/proc/1/ns/net')  # go back to default netns
+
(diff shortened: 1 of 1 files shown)


pyroute2/netns/__init__.py
```diff
@@ -222,6 +222,11 @@ def setns(netns, flags=os.O_CREAT, libc=None):
         - O_CREAT -- create netns, if doesn't exist
         - O_CREAT | O_EXCL -- create only if doesn't exist
 
+    Note that "main" netns as no name. But you can access it with::
```
svinota: … has no name

svinota:
Another way is to open `/proc/self/ns/net`, save the fd, and switch back with `setns()`. This allows to revert to own namespace even if it was not the main one.

svinota:
Probably I have to dump it also to the docstring.

svinota:
As an idea, looks at the commit — e7810d93cc330e35cd3fba38952b683cf0e0994b

ffourcot:
Thank you!

svinota:
Thank you for the issue. I had to implement it a long time ago, but didn't.

PR #591, opened 2019-03-05
Skbinfo in wiset

cdaymand:
Hello,

I added the support for skbinfo in the wiset ipset helper.
You can add entries with skbmark as tuple like for ipset or as a string as in the ipset command documentation.
I also added one test in the test_wiset.py file.

Best regards

Charles DAYMAND


svinota:
Super. Thanks a lot!

PR #592, opened 2019-03-05
Implement nftables rules expressions

inemajo:


svinota:
Oh my. That will take a day to review :)

Thanks for such fantastic work.

tests/general/test_nftables.py
```diff
@@ -0,0 +1,50 @@
...
+from pyroute2.nftables.rule import NFTRule
+
+from utils import require_user
+
+#NFT_BIN_PATH = "/root/nft/nftables/src/nft"
+NFT_BIN_PATH = "nft"
+NS_NAME = 'pyroute2_test_nftable'
+
+
+class NFTables_test(object):
```
svinota: I'm going to disable this test for a while since the nft command does not support `export json` anymore, and `-j list ruleset` provides a completely different output.

We have to rework these tests somehow.

PR #597, opened 2019-04-03
Wiset replace update

cdaymand:
Hello,

This pull request add the support of kwargs list for the replace_entries method of wiset in addition to a basestring list.
This allow to use this "atomic" operation to replace the content of an ipset having skbinfo or comments
I also added a test for it.

Best regards

Charles


svinota:
Looks good. Thanks!

PR #603, opened 2019-05-14
Fix seg6 and seg6local.End.DT4

cscarpitta:


pyroute2/netlink/rtnl/rtmsg.py
```diff
@@ -255,7 +255,7 @@ def encode(self):
                     # Iterate over segments
                     for seg in segs:
                         # Convert to network byte order and add to value
-                        self['segs'] += inet_pton(family, seg)
+                        self['segs'] += inet_pton(AF_INET6, seg)
```
svinota: Is it true for `family == AF_INET` ?

pyroute2/netlink/rtnl/rtmsg.py
```diff
@@ -214,8 +214,8 @@ def get_family(self):
             def encode(self):
                 # Retrieve the family
                 family = self.get_family()
-                # Seg6 can be applied only to IPv6
-                if family == AF_INET6:
+                # Seg6 can be applied only to IPv6 and IPv4
```
svinota: May I ask you to give some examples in the docs also, like for SEG6?

pyroute2/netlink/rtnl/rtmsg.py
cscarpitta: Yes, my commit adds support for seg6 ip4ip6 encapsulation
The linux kernel already supports this feature starting from the 4.14 version
The iproute2 suite already implements a command for ip4ip6 encapsulation

seg6 ip4ip6 encapsulation was introduced with the following commit:
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=32d99d0b670299720dd0db92a974c9612c230889

The destination route is an AF_INET6 or AF_INET address
In both cases the segments in the segment routing header belong to the AF_INET6 family

PR #605, opened 2019-05-20
Fix IPDB + NetNS usage example to also close() NetNS instance after use

### LINKED ISSUE #539: By using IPDB context manager he don`t close NetNS() instance.
Hi, I notice that context manager do not close NetNS() instance when exit.

    ns = NetNS("ns1")
    with IPDB(nl=ns) as ip:
        pass
    print(ns.closed) #==> False

Regards.

jkjuopperi:
Closes #539


svinota:
Thanks!

PR #611, opened 2019-06-15
add HT and VHT decode

linuxlizard:
start of architecture to decode the rest of the IEs

I'm using a class IE as a base for the Information Elements' decoded. The IEs will contain a list of named tuples with 'name','value' as fields. The IEs' values can be nested. The values will be grouped according to the IEEE80211-2016 spec.

### DIFF
--- a/examples/nl80211_scan_dump.py
+++ b/examples/nl80211_scan_dump.py
@@ -14,60 +14,175 @@ from pyroute2.netlink.nl80211 import NL80211_NAMES
 logging.basicConfig(level=logging.DEBUG)
 
 logger = logging.getLogger("scandump")
-logger.setLevel(level=logging.INFO)
+logger.setLevel(level=logging.DEBUG)
+# logger.setLevel(level=logging.INFO)
 
-# interface name to dump scan results
-ifname = sys.argv[1]
+logging.getLogger("pyroute2").setLevel(level=logging.DEBUG)
 
-iw = IW()
 
-ip = IPRoute()
-ifindex = ip.link_lookup(ifname=ifname)[0]
-ip.close()
+def print_ht_capabilities(ht_capa):
+    if not ht_capa:
+        return
 
-# CMD_GET_SCAN doesn't require root privileges.
-# Can use 'nmcli device wifi' or 'nmcli d w' to trigger a scan which will fill
-# the scan results cache for ~30 seconds.
-# See also 'iw dev $yourdev scan dump'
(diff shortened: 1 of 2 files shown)


linuxlizard:
I watched the Travis CI output. I'll start cleaning up the warnings in nl80211_scan_dump.py

svinota:
Thanks! To be merged asap

linuxlizard:
What would you think about something like the following instead?  It's using an integer -> class map to convert the IE into a specific class instance. Also the decoded value[] is now integer based instead of string.

My eventual goal is to get the scan results as a JSON dump. From a JSON dump, I can create a Flask app that serves the JSON. And ta-da! I have a remote wifi scanner.

```
           def decode(self):
                nla_base.decode(self)

                self.value = {}

                init = offset = self.offset + 4

                ie_class_map = {
                    NL80211_BSS_ELEMENTS_HT_CAPABILITIES : HT_Capabilities,
                    NL80211_BSS_ELEMENTS_HT_OPERATION : HT_Operation,
                    NL80211_BSS_ELEMENTS_EXT_CAPABILITIES : Extended_Capabilities,
                    NL80211_BSS_ELEMENTS_VHT_CAPABILITIES : VHT_Capabilities,
                    NL80211_BSS_ELEMENTS_VHT_OPERATION : VHT_Operation,
                }

                while (offset - init) < (self.length - 4):
                    (msg_type, length) = struct.unpack_from('BB',
                                                            self.data,
                                                            offset)
                    try:
                        cls = ie_class_map[msg_type]
                    except KeyError:
                        pass
                    else:
                        self.value[msg_type] = cls(self.data[offset:offset + length + 1])
                        self.value[msg_type].decode()
                        offset += length + 2

```

svinota:
Looks nice, you're right. So, I'll wait then for the PR and merge.

Thanks!

linuxlizard:
Quick update: still working on this. Been on vacation with family the last week so I'm a little distracted. I have first steps of my JSON output working. I'm worried about breaking previous nl80211 data formats. I'm learning how to use your CI so I can test before adding to the PR.

svinota:
The nl80211 part is not heavily tested, unfortunately, so you may expect surprises there. If so — please don't hesitate to ping me, we can run debugging sessions together after some time planning. Also ping me in case of any questions.

linuxlizard:
Too many merge conflicts with rebase -i so going to create a new PR.

PR #615, opened 2019-06-23
Add support for IPoIB interface

adrianchiris:
- Add ipoib nla to allow creation/query of ipoib interfaces
  with the defined attributes.

- Document ipoib usage in iproute.linux.RTNL_API.link() docstring.


svinota:
Thanks!

PR #627, opened 2019-08-08
Set infiniband port and node GUID

Hamdy-khader:
This patch adds the abbility to configure infiniband port GUID
and node GUID by implementing classes `ib_port_guid` and `ib_node_guid`
in class IFLA_VF_INFO for serializing and deserializing GUID.


svinota:
Thanks!

PR #632, opened 2019-08-27
Change 32hex conversion functions to support both py2 and py3.

liam-mcb:
This fixes the `from_dict()` and `to_dict()` methods on Python3.
They previously gave the following errors.
```
  File "/home/liam/work/pyroute2/pyroute2/nftables/parser/expr.py", line 109, in from_32hex
    res = chr(int(val[i:i + 2], 16)) + res
TypeError: can only concatenate str (not "bytes") to str
```
and...
```
  File "/home/liam/work/pyroute2/pyroute2/nftables/parser/expr.py", line 130, in to_32hex
    res += format(ord(c), '02x')[::-1]
TypeError: ord() expected string of length 1, but int found
```

### DIFF
--- a/pyroute2/nftables/parser/expr.py
+++ b/pyroute2/nftables/parser/expr.py
@@ -103,11 +103,7 @@ class NFTData(object):
     @classmethod
     def from_dict(cls, data):
         def from_32hex(val):
-            val = val[2:]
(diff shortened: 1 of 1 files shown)


svinota:
thanks!

PR #636, opened 2019-09-27
rtnl: fix type of RTA_PREF

### LINKED ISSUE #630: IPRoute().flush_routes warnings in kernel log when RTA_PREF present
In `flush_routes`, if `get_routes` returns with `RTA_PREF`, the put to `RTM_DELROUTE` results in the following dmesg, at least on Linux 5.0.0:
```
netlink: 'python3': attribute type 20 has an invalid length.
```
The routes *are* removed, but this causes future route modifications to fail. Removing `RTA_PREF` from `route['attrs']` fixes the issue.

PaulSD:
Fixes #630


svinota:
Thanks!

PR #639, opened 2019-10-07
Add support route('append', ...) operation

crosser:
iproute2 supports `ip route append ...` command, that allows, possibly
among other things, to set up IPv6 multipath 'device only' routes that
cannot be confitured via the kernel "multipath" interface. Relevant code
in the iproute2 source in ip/iproute.c:2295 looks like this:

        if (matches(*argv, "append") == 0)
                return iproute_modify(RTM_NEWROUTE, NLM_F_CREATE|NLM_F_APPEND,
                                      argc-1, argv+1);

This commit aims to replicate this functionality in pyroute2.

Signed-off-by: Eugene Crosser <evgenii.cherkashin@cloud.ionos.com>

### DIFF
--- a/pyroute2/iproute/linux.py
+++ b/pyroute2/iproute/linux.py
@@ -1761,11 +1761,12 @@ class RTNL_API(object):
                             'srh': {'segs': '2000::5,2000::6',
                                     'hmac':0xf}})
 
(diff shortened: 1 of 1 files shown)


svinota:
Thanks!

PR #645, opened 2019-10-28
Added IPset example

jeroenvermeulen:
Added IPset example to save another newbie some time

### DIFF
--- a/examples/ipset.py
+++ b/examples/ipset.py
@@ -1,0 +1,39 @@
+from pyroute2.ipset import IPSet, PortRange, PortEntry
+
+ipset = IPSet()
+ipset.create("foo", stype="hash:ip")
+ipset.add("foo", "198.51.100.1", etype="ip")
+ipset.add("foo", "198.51.100.2", etype="ip")
+print ipset.test("foo", "198.51.100.1")   # True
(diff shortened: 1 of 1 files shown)


ffourcot:
Hello,

There are already some examples with bitmap_ports_range and getprotobyname in documentation, at the end of this section:

https://docs.pyroute2.org/ipset.html#pyroute2.ipset.IPSet.add

Is it not enough? I was not aware of the examples folder, to be honest.

By the way, WiSet module is more high level to manage ipset, and probably (I hope) easier to use. In all cases, it's a good idea to improve documentation/examples coverage of this part, thanks to pointing this out.

jeroenvermeulen:
Wan't aware of the documentation, did't reach the end of the [README](https://github.com/svinota/pyroute2/blob/master/README.md), maybe you should mention it somewhere in the beginning. 
The examples I did find were only explaining adding to IPset, wouldn't do any harm to merge my PR, would it?

ffourcot:
Sure, it does not hurt. I will let @svinota to have a look on it.

I was just curious of missing part of documentation, and how to improve it. Your suggestion to add a better link to documentation from README is probably a good point.

ffourcot:
Merged, thanks!

PR #648, opened 2019-11-11
Add lladdr ipv4 support

mikezsin:


YarickR:
I'd like to elaborate on this pull request: 
sometimes (with multipoint GRE tunnels, f.e.) link-local address of a peer is an IP , instead of MAC; 
using l2addr in this case makes manipulating neighbour records impossible. 
this patch allows adding and removing GRE-connected neighbours. I haven't tested it with regular L2 records, though, and I'm not sure this patch is entirely correct, but please take a look, maybe you'll have better idea how to fix the issue.

pyroute2/netlink/__init__.py
```diff
@@ -1823,6 +1823,36 @@ def decode(self):
...
+        '''
+        Decode MAC or IPv4 address.
+        '''
+
+        __slots__ = ()
+        sql_type = 'TEXT'
+        fields = [('value', 's')]
+
+        def encode(self):
+            if self.value.find(':') > -1:
```
ffourcot: Did you tried with IPv6? I'm afraid that it will be detected as a MAC address

pyroute2/netlink/__init__.py
```diff
@@ -1823,6 +1823,36 @@ def decode(self):
...
+                raise TypeError('Unsupported value {}'.format(self.value))
+            nla_base.encode(self)
+
+        def decode(self):
+            nla_base.decode(self)
+            if self.length == 6: 
+                self.value = ':'.join('%02x' % (i) for i in
+                                  struct.unpack('BBBBBB', self['value']))
+            elif self.length == 4:
+                self.value = inet_ntop(family, self['value'])
```
ffourcot: family is undefined here?

pyroute2/netlink/__init__.py
```diff
@@ -1823,6 +1823,36 @@ def decode(self):
...
+        __slots__ = ()
+        sql_type = 'TEXT'
+        fields = [('value', 's')]
+
+        def encode(self):
+            if self.value.find(':') > -1:
+                self['value'] = struct.pack('BBBBBB',
+                                            *[int(i, 16) for i in 
+                                              self.value.split(':')])
+            elif self.value.find('.') > -1:
```
ffourcot: In my opinion, decoding values based on length is OK. But for encoding, we should wait explicit values from user, and not guessing values. Since encoding is done in python, we could set the type somewhere and use the proper function to encode data

pyroute2/netlink/__init__.py
```diff
@@ -1823,6 +1823,36 @@ def decode(self):
...
+            nla_base.encode(self)
+
+        def decode(self):
+            nla_base.decode(self)
+            if self.length == 6: 
+                self.value = ':'.join('%02x' % (i) for i in
+                                  struct.unpack('BBBBBB', self['value']))
+            elif self.length == 4:
+                self.value = inet_ntop(family, self['value'])     
+            else:
```
ffourcot: Add IPv6 support here?

PR #652, opened 2019-11-28
WireGuard: multiple fixes

Vic063:
Since the update of the WireGuard code now using .NLA_EXACT_LEN attributes (see https://github.com/WireGuard/WireGuard/blob/19b2fb70620d5d664549b1e9dd5f123ab1a1baf5/src/netlink.c#L32), header types of the following classes are not correctly set when encoding NLAs:
- wgpeer_a_allowedips
- wgpeer_allowedip
- wgdevice_a_peers
- wgdevice_peer

So I forced them, taken from dumps of netlink messages using the 'wg' command.
I don't know if it's really related with .NLA_EXACT_LEN attributes but I did not understand how these values are built.

It also includes a fix for Python2 and empty IPv6 endpoint address.


Vic063:
Sorry, I was too pressed and forget to rebase from master...
New PR is here: #653

PR #653, opened 2019-11-28
WireGuard: multiple fixes

Vic063:
Since the update of the WireGuard code now using .NLA_EXACT_LEN attributes (see https://github.com/WireGuard/WireGuard/blob/19b2fb70620d5d664549b1e9dd5f123ab1a1baf5/src/netlink.c#L32), header types of the following classes are not correctly set when encoding NLAs:

wgpeer_a_allowedips
wgpeer_allowedip
wgdevice_a_peers
wgdevice_peer
So I forced them, taken from dumps of netlink messages using the 'wg' command.
I don't know if it's really related with .NLA_EXACT_LEN attributes but I did not understand how these values are built.

It also includes a fix for Python2 and empty IPv6 endpoint address.

### DIFF
--- a/pyroute2/netlink/generic/wireguard.py
+++ b/pyroute2/netlink/generic/wireguard.py
@@ -76,9 +76,10 @@ from time import ctime
 
 from pyroute2.netlink import genlmsg
 from pyroute2.netlink import nla
-from pyroute2.netlink import NLM_F_REQUEST
 from pyroute2.netlink import NLM_F_ACK
 from pyroute2.netlink import NLM_F_DUMP
+from pyroute2.netlink import NLA_F_NESTED
+from pyroute2.netlink import NLM_F_REQUEST
 from pyroute2.netlink.generic import GenericNetlinkSocket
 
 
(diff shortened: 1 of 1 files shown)


pyroute2/netlink/generic/wireguard.py
```diff
@@ -152,6 +158,10 @@ class wgdevice_peer(nla):
                        ('WGPEER_A_ALLOWEDIPS', 'wgpeer_a_allowedips'),
                        ('WGPEER_A_PROTOCOL_VERSION', 'uint32'))
 
+            def encode(self):
+                self['header']['type'] = 0x8000
```
ffourcot: Can you try to set `nla_flags = NLA_F_NESTED` instead to overwrite encode here?

pyroute2/netlink/generic/wireguard.py
```diff
@@ -189,7 +199,7 @@ def encode(self):
                         self['family'] = AF_INET
                         self['addr4'] = unpack('>I',
                                                inet_aton(self['addr']))[0]
-                        self['addr6'] = bytearray(4)  # Set 4 bytes to NULL
+                        self['addr6'] = b'\x00\x00\x00\x00\x00\x00\x00\x00'
```
ffourcot: Previous code was not working?

Vic063: It worked but I think NLA auto-padding was our guardian angel.

pyroute2/netlink/generic/wireguard.py
Vic063: You're right. Force pushed here d82533771a43ea5debca57adf5ab2e1a492098b1

pyroute2/netlink/generic/wireguard.py
Vic063: Tested with both IPv4 and IPv6 addresses on both Python versions. It works properly.

Issue #646, opened 2019-10-29
filter api can not match mac address


JackSunshine:
pyroute2 filter API doesn't support matching mac address, for example, I can install a filter rule through the following command line:

> # tc filter add dev vnet1 parent 1:0 protocol ip prio 5 u32 match u16 0x0800 0xffff at -2 match u16 0xca67 0xffff at -4 match u32 0x5254005c 0xffffffff at -8  flowid 1:10

but I can't install this rule through the tc API, cause the offset is negative.

svinota:
Let's see what can we do here. Thanks for the report

Vic063:
@JackSunshine: Could you please test the fix issued in PR #656 ?

Here is the code I used to test:
from pyroute2 import IPRoute, protocols

ipr = IPRoute()
idx = ipr.link_lookup(ifname='vnet1')[0]
ipr.tc('add-filter', 'u32', idx, parent=0x10000, prio=5, protocol=protocols.ETH_P_IP, target=0x10010, keys=["0x0800/0xffff-2", "0xca67/0xffff-4", "0x5254005c/0xffffffff-8"])

PR #656, opened 2019-12-13
[WIP] u32: fix for negative offsets

Vic063:
This fix should resolve the issue #646

### DIFF
--- a/pyroute2/netlink/rtnl/tcmsg/cls_u32.py
+++ b/pyroute2/netlink/rtnl/tcmsg/cls_u32.py
@@ -159,14 +159,20 @@ class options(nla, nla_plus_police):
                 return (new_key, field)
 
             # 'header' array to pack keys to
-            header = [(0, 0) for i in range(256)]
+            header = [(0, 0) for i in range(-256, 256)]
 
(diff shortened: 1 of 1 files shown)


svinota:
Sounds reasonable, so I would merge it one day when the [WIP] will be cleared.

pyroute2/netlink/rtnl/tcmsg/cls_u32.py
```diff
@@ -159,14 +159,20 @@ def cut_field(key, separator):
...
+                if key.find('-') != -1:
+                  (key, offset) = cut_field(key, '-')
+                  use_minus = True
+                else:
+                  (key, offset) = cut_field(key, '+')
+                  use_minus = False
+
                 offset = int(offset, 0)
                 # a little trick: if you provide /00ff+8, that
                 # really means /ff+9, so we should take it into
```
reshmasreekumar, line 178: @Vic063 you will have to change the logic of updating offset at line number 185 .i.e.

offset = offset + 1 if not use_minus else offset - 1

svinota:
Thanks, trying.

PR #657, opened 2019-12-16
Enable action parameter for basic, flow and fw filters

Vic063:
'action' support was not enable on these three filters. This patch enables it.

### DIFF
--- a/pyroute2/netlink/rtnl/tcmsg/cls_fw.py
+++ b/pyroute2/netlink/rtnl/tcmsg/cls_fw.py
@@ -4,6 +4,9 @@ from pyroute2.netlink import nla
 from pyroute2.netlink.rtnl.tcmsg.act_police import nla_plus_police
 from pyroute2.netlink.rtnl.tcmsg.act_police import get_parameters \
     as ap_parameters
+from pyroute2.netlink.rtnl.tcmsg.common import TCA_ACT_MAX_PRIO
+from pyroute2.netlink.rtnl.tcmsg.common_act import get_tca_action
+from pyroute2.netlink.rtnl.tcmsg.common_act import nla_plus_tca_act_opt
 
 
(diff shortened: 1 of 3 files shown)


pyroute2/netlink/rtnl/tcmsg/cls_fw.py
```diff
@@ -38,5 +43,16 @@ class options(nla, nla_plus_police):
                ('TCA_FW_CLASSID', 'uint32'),
                ('TCA_FW_POLICE', 'police'),  # TODO string?
                ('TCA_FW_INDEV', 'hex'),  # TODO string
-               ('TCA_FW_ACT', 'hex'),  # TODO
+               ('TCA_FW_ACT', 'tca_act_prio'),
                ('TCA_FW_MASK', 'uint32'))
+
+    class tca_act_prio(nla):
```
ffourcot: We should try to not duplicate this code everywhere for each filter. We could perhaps introduce a common ancestor class with this kind of code?

Vic063: After looking at the code and made some experiments with code factorization, it seems to be hard to do what you are looking for.

Look at the function 'compile_nla' in 'netlink/__init__.py' file: sadly, it looks up the class by its name using 'getattr(**self**, "tca_act_prio")' so it will never find the declaration of the 'tca_act_prio' outside of the current class.

Moreover, it might be due to my knowledge limitations of Python but I don't know if it will be easy to override sub-classes declared inside 'tca_act_prio' class. For reference, I invite you to compare 'action' declarations of sched_bpf.py, act_fw.py and cls_u32.py.
Each declaration of the 'tca_act_prio' class is different from the others.

We can talk about it tomorrow.

ffourcot:
Merged, thanks!

PR #660, opened 2020-01-15
Conntrack filter

inemajo:
This pull request add a filter fonctionnality for the Conntrack().dump() method.

We are working on kernel side filtering as well, see https://marc.info/?l=netfilter-devel&m=157675221817672&w=2

We will add pyroute2 support as soon as it is merged upstream.


svinota:
Thanks, merged! Some pep8 fixes will follow.

PR #662, opened 2020-01-20
cake: add stats_app decoder

Vic063:
This commit allows to get stats from a cake filter including stats per tin.


svinota:
Thanks!

PR #663, opened 2020-01-27
Wildcard and physdev in hash:net,iface

ffourcot:
Hello,

Some commits to support new "wildcard" matching support, added in kernel 5.5. I didn't add tests for it now, since it needs a very recent kernel. But I can add it if needed.

Architecture is the same than old "physdev" feature, that I added in same time (with tests).


svinota:
Yes, please add the test as a separate PR, I'll make the test conditional.

PR #666, opened 2020-02-06
wireguard: use nla arrays

svinota:
In order to simplify the code the nla arrays encoding has been fixed using nla_flags for array members as well.


svinota:
@Vic063 , I would like you to review the code as well before I merge it.

And thanks a lot for the previous fixes.

Vic063:
@svinota, I will look at it this afternoon ;)

PR #671, opened 2020-02-07
tc/meta: add meta support

Vic063:
This patch adds meta ematch support for basic filter.


svinota:
I plan to merge the code, and tomorrow propose some changes for your review. We can simplify a bit it in two different ways.

PR #675, opened 2020-02-12
Ethtool

inemajo:
Implementation of ethtool for pyroute2, using kernel ioctl & netlink API.


svinota:
Yep, here it is. Thanks!

PR #679, opened 2020-02-21
Add psutil dependency for cli

jonringer:
the ss2 cli script imports it:

```
$ ./results/python37Packages.pyroute2/bin/ss2
  File "/nix/store/7wsicyf6lf5z0fj8flxi5kkjg2pv1prf-python3.7-pyroute2-0.5.9/bin/.ss2-wrapped", line 28, in <module>
    import psutil
ModuleNotFoundError: No module named 'psutil'
```
https://github.com/svinota/pyroute2/blob/7d1a19c60aeceb3851413dcdcd547ee2f254bc11/cli/ss2#L26


jonringer:
related Packaging PR https://github.com/NixOS/nixpkgs/pull/80691

setup.py
```diff
@@ -33,6 +33,7 @@
       author_email='peter@svinota.eu',
       url='https://github.com/svinota/pyroute2',
       license='dual license GPLv2+ and Apache v2',
+      install_reqiures=['psutil~=5.0'],
```
Mic92:
```suggestion
      install_requires=['psutil~=5.0'],
```

svinota:
I don't know how I missed this dependency merging ss2 :|

Give me a day or two to figure out how to avoid this dependency at all.

Mic92:
You would first need to read `/proc/net/*` to get all the inodes and than map those back to processes.

svinota:
@jonringer @Mic92 Would it work to make `ss2` an optional component for some time until I strip psutil code from the script? As it is here: https://github.com/svinota/pyroute2/commit/e4416a38d63e180ebdc13a3ceea2e3813b2f4901

There are too many users that will not be happy to have psutil in the requirements, so I can not just force to install it wildly together with the library.

setup.py
jonringer: there could be a breaking change on the 6.0 release, we would want an upper bound

Mic92:
I suppose that would work too. In nixpkgs it would be an all or because we tend to avoid optional dependencies. So we would either drop the script for the time being or have psutil.

jonringer:
So I'm not the maintainer of the pyroute2 packages, I just noticed this discrepancy when reviewing a package bump.

Thought upstream would like to know.

If you want to vendor the code like @Mic92 mentioned https://github.com/svinota/pyroute2/pull/679#issuecomment-589601548 or do another alternative is up to you.

> There are too many users that will not be happy to have psutil in the requirements, so I can not just force to install it wildly together with the library.

Then @Mic92's  recommendation would probably be best, unless you want to remove the related functionality

cherusk:
ss2 - gold and silver!
@svinota , as discussed offline, let's make ss2 an extra for now.

jonringer:
superseded by https://github.com/svinota/pyroute2/pull/680

PR #680, opened 2020-02-21
ss2: make an extra

svinota:
Declare psutil in `extras_require` and raise an exception if it is
not present in the system

Bug-Url: https://github.com/svinota/pyroute2/pull/679

### DIFF
--- a/cli/ss2
+++ b/cli/ss2
@@ -23,7 +23,10 @@ import argparse
 from socket import (AF_INET,
                     AF_UNIX
                     )
-import psutil
+try:
+    import psutil
+except ImportError:
+    psutil = None
 from pyroute2 import DiagSocket
 from pyroute2.netlink.diag import (SS_ESTABLISHED,
                                    SS_SYN_SENT,
@@ -552,6 +555,9 @@ def prepare_args():
 
 def run(args=None):
 
+    if psutil is None:
+        raise RuntimeError('ss2 requires python-psutil >= 5.0 to run')
+
     if not args:
         args = prepare_args()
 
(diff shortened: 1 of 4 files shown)


svinota:
@ffourcot @jonringer 

please review this change before I proceed — here I change distutils to setuptools by default, and it may affect

cli/ss2
```diff
@@ -578,7 +584,7 @@ def run(args=None):
     with DiagSocket() as ds:
         ds.bind()
         for p in protocols:
-            p(ds, args, _user_ctxt_map)
+                p(ds, args, _user_ctxt_map)
```
cherusk: lapsus!

cherusk:
Should we also mention it in some readme local to  cli/, but optional.

cli/ss2
svinota: indeed

jonringer: this looks over-indented. You use 4-wide everywhere, but 8 here

setup.py
```diff
@@ -66,6 +66,7 @@
                 'pyroute2.remote'],
       scripts=['./cli/ss2',
                './cli/pyroute2-cli'],
+      extras_require={'ss2': ['psutil>=5.0']},
```
jonringer:
```suggestion
      extras_require={'ss2': ['psutil~=5.0']},
```
you will still want an upperbound, as 6.0 may break your use cases (~=5.0 -> >=5,<6)

cli/ss2
svinota: fixed

PR #689, opened 2020-03-23
Add lladdr ipv4 and ipv6 support for neighbor

thbtcllt:
This pull request is an improvement of #648 that is still pending

This version supports ipv6 contrary to #648 

For encoding we accept three formats whatever the expectation of the interface. This default is also present in ip route. For example if ze provide a mac address as link layer for a gre4 interface the kernel will translate it as a ipv4 address
ip -4 tunnel add gre1 mode gre local 10.125.0.1 dev eth0 ttl 64
ip address add 192.168.0.1/24 dev gre1
ip link set dev gre1 up
ip neigh replace 192.168.0.2 lladdr 52:53:54:55:56:57 dev gre1

ip neigh show
192.168.0.2 dev gre1 lladdr 82.83.84.85 PERMANENT

in fact the link layer address is a binary byte array up to 32 characters. Only the interface knows the expected size. Retrieve this information through the ifindex is too complex and overkill (not check by ip route)


svinota:
Excellent! Thank you

PR #692, opened 2020-03-26
Ethtool features fixes

inemajo:
Hello, there was some littles bug on ethtool for get/set features, theses patches correct them


svinota:
Thanks!

PR #695, opened 2020-04-07
Enable generator on Conntrack() and reduce memory usage

ffourcot:
Hello,

Conntrack().dump_entries() return a lot of objects, and is consuming huge memory (on our tests, more than 200Mo for 17000 entries, for example).

Thanks to nlm_generator now by default to True (side note : more documentation on this configuration could be great!) our script to dump and filter conntrack entries consumes less than 50Mo.

It has less performance impacts, but we also adds some __slots__ definitions on Conntrack related objects.

Conntrack() class is still compatible with nlm_generator sets on False.

### DIFF
--- a/pyroute2/conntrack.py
+++ b/pyroute2/conntrack.py
@@ -86,6 +86,8 @@ class Conntrack(NFCTSocket):
     """
     High level conntrack functions
     """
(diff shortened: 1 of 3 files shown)


svinota:
Thanks!

PR #705, opened 2020-05-12
Add support for 'replace' command in addr method

Shushuda:
addr method only accepts 'add', 'delete' and 'dump' commands.
This commit adds 'replace' and a test.


svinota:
Thanks!

Issue #562, opened 2018-11-12
"address is not allocated"


tomkcook:
My code calls `IPRoute.get_addr(family=socket.AF_INET, label='eth0')` fairly frequently.  After a bit, it starts doing this:

```
Traceback (most recent call last):
  File "InterfaceUtils.py", line 336, in valid_ip4_address
  File "pyroute2/iproute/linux.py", line 303, in get_addr
  File "pyroute2/iproute/linux.py", line 1271, in addr
  File "pyroute2/netlink/nlsocket.py", line 352, in nlm_request
  File "pyroute2/netlink/nlsocket.py", line 830, in nlm_request
  File "pyroute2/common.py", line 407, in alloc
  File "pyroute2/common.py", line 483, in free
KeyError: 'address is not allocated'
```

I'm not clear what's going on here - any suggestions?

tomkcook:
This is a threading issue.  It can be readily reproduced with the following code:

```
import threading

def test():
    from pyroute2 import IPRoute
    import socket

    ipr = IPRoute()
    def test_thread():
        for j in range(10000):
            details = []
            for i in range(10000):
                ipr.get_addr(family=socket.AF_INET, label='wlp1s0')

    for i in range(10):
        threading.Thread(target=test_thread).start()

test()
```

Moving `ipr = IPRoute()` two lines down, so that there is one such object for each thread, eliminates the problem.

I will try to put together a pull request that addresses this tomorrow.

tomkcook:
Actually, I've decided to fix this in my code by making my instances of the `IPRoute` and `IW` objects thread-local, sorry.

svet-b:
Just wanted to say that I encountered the same issue. Wish that the error message was more descriptive, as it really sent me on quite a wild goose chase. And unfortunately this issue page for some reason does not come up when googling " pyroute2 KeyError 'address is not allocated' ".

Indeed, it was solved by not having a single IPRoute object that's used across threads.

svinota:
IPRoute objects must be thread-safe, so if they fail to operate this way — it's a bug. I reopen the issue. @svet-b thanks for bringing it up.

PR #706, opened 2020-05-13
Fix double free of msg_seq in nlm_request

chrspeich:
This PR does two things.

 1. First it fixes the `address not allocated` error described in #562 
 2. It changes the `print` for the retry error to a `log.warning`.

While I don't quite understand why the `address not allocated` error does not always happen when the request is retried, this fixes the problem.


svinota:
Thanks, testing!

Issue #700, opened 2020-04-16
make install fails on README.md


jtluka:
With version 1b5131a24cfe9d623151bcb0190235e9f729a036 I am not able to install the pyroute through make install (or python3 setup.py install).

I was able to workaround this with the following patch:
```
diff --git a/setup.py b/setup.py
index 5c387f01..b0f1449e 100644
--- a/setup.py
+++ b/setup.py
@@ -23,7 +23,7 @@ module = __import__(config.get('setup', 'setuplib'),
                     ['setup'], 0)
 setup = getattr(module, 'setup')
 
-readme = open("README.md", "r")
+readme = open("README.rst", "r")
 
 
 setup(name='pyroute2',
```

svinota:
Thanks! To be fixed tonight

PR #710, opened 2020-05-25
docs: Fix README.rst path in setup.py

kaechele:
The change in e5f610de7e92782f6d66fd3f53baf1df6bdacc88 introduced a
README.rst in place of the README.md but this was not reflected in
setup.py.

Signed-off-by: Felix Kaechele <felix@kaechele.ca>

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -23,7 +23,7 @@ module = __import__(config.get('setup', 'setuplib'),
                     ['setup'], 0)
 setup = getattr(module, 'setup')
 
-readme = open("README.md", "r")
+readme = open("README.rst", "r")
 
 
 setup(name='pyroute2',


kaechele:
I saw something was done about it in #700 but that doesn't fix installing through setup.py (e.g. using pip with a git repository rather than through PyPI).

svinota:
Using the `rst` version as the README breaks formatting on pypi web page :|

Thanks for the heads-up anyway, I'll think what we can do with that.

jacob-keller:
I ran into this recently and had to modify setup.py to get it to work.

svinota:
Should be fixed by https://github.com/svinota/pyroute2/commit/ab278ffd74623e24f5979c9b3cc62f51a1150581

Issue #724, opened 2020-07-28
IPRoute.link_lookup(address='fa:16:3e:92:6b:39') returns [] on 0.5.13


johnsom:
On 0.5.13 link_lookup using an address only returns [] where on 0.5.12 the same lookup returns the proper index.

To reproduce on 0.5.13:
```
import pyroute2
ipr = pyroute2.IPRoute()
ipr.link_lookup(address='<valid MAC>')
```
Result:
[]

No matter what MAC I put in, I always get []

On 0.5.12:
```
import pyroute2
ipr = pyroute2.IPRoute()
ipr.link_lookup(address='<valid MAC>')
```
Result:
[14]

This issue has broken OpenStack Octavia.

johnsom:
In this patch: https://github.com/svinota/pyroute2/commit/3e0b8bd2a34a188f41c465d5313a5dc9fee8db2f

if list(kwarg) <= ['index', 'ifname']:

evaluates to:
>>> ['address'] <= ['index', 'ifname']
True

Which then leads to this exception:
(22, 'Invalid argument')
and [] always being returned.

ffourcot:
Since I'm filling a little bit guilty, I made a small proposal to fix this regression.

It's a side effect of https://github.com/svinota/pyroute2/issues/712

johnsom:
I can confirm that this proposed patch resolves our use case.

If this is an acceptable patch for all, can we get a release that includes it soon?

smurfix:
Shouldn't this issue be closed?

johnsom:
I agree, 0.5.14 resolves the issue, so I will click the closed button. Apologies if I am not following the process on this project.

Issue #712, opened 2020-06-02
IPRoute.link_lookup() is too slow


ffourcot:
Hello,

link_lookup is mandatory for many use cases since we need interface index for almost all change/update.

However, current pyroute2 implementation is too slow. Especially in use case with many interfaces. On a computer with about thousands interfaces, we got:

```
In [19]: ipr = IPRoute()
In [20]: %timeit ipr.link_lookup(ifname="lo")
1 loop, best of 5: 294 ms per loop
```
When with another implementation:

```
In [22]: %timeit ipr.link("get", ifname="lo")[0]["index"]
1000 loops, best of 5: 491 µs per loop
```

In link_lookup, all filtering is done in userspace, since "dump" does not support kernel side filtering. With a simple "get" (kernel supports filtering on it), performance improvement is huge

svinota:
Reasonable. Ok, thanks, working on the improvement.

svinota:
Should be a bit better now.

PR #725, opened 2020-07-28
iproute: fix link_lookup when index/ifname is not set

ffourcot:
Use set intersection to check if index nor ifname is in kwargs. And add
a trivial test for this

Bug-Url: https://github.com/svinota/pyroute2/issues/724
Related-To: https://github.com/svinota/pyroute2/issues/712


johnsom:
Confirmed that this fix resolves the issue we saw and reported in 724.

rm-you:
Yeah, we really need this to unblock our gates. Hopefully this can be reviewed and merged soon! 👍

PR #729, opened 2020-08-07
devlink: add new commands and attributes

jacob-keller:
The support for devlink commands in pyroute2 lacks many recent additions
to the devlink generic netlink interface. Add them.

Signed-off-by: Jacob Keller <jacob.e.keller@intel.com>


jacob-keller:
I am not 100% convinced I got all of the attributes correct, as I wasn't sure how to properly handle nested NLAs.

jacob-keller:
I need to use *recursive to enable this because devlink attributes are defined recursively. I also noticed that the parser is unable to handle multiple depths of nested NLA attributes

jacob-keller:
> I need to use *recursive to enable this because devlink attributes are defined recursively. I also noticed that the parser is unable to handle multiple depths of nested NLA attributes

Oh.. that might be because several of the devlink nested attributes weren't properly marked with NLA_F_NESTED... (nla_nest_start_noflag in the kernel code...)

jacob-keller:
So it seems like this is an issue because using *recursive makes it break when trying to parse nested NLA attributes. (It assumes that there is a header available with the genlmsg header format).. but that won't be there for the nested NLAs.

I'm really not sure how to resolve this: the devlink attributes are all defined in one giant block, and trying to figure out which ones can be valid to nest where and separate them into separate classes seems quite difficult.

I ended up also trying a subclass "devlink(nla)" with identical attribute definition as the main class, but this seems to still be unable to parse the nested attributes of DEVLINK_ATTR_INFO_VERSION_* attributes.

PR #741, opened 2020-10-15
wiset: Adding 'port' attribute for key of WiSet().content

inemajo:
Hello,

This commit permit to format the key of the WiSet().content with attributes of port like:

{'1.2.3.4/32,tcp:60005': IPStats(...) ...

It add "tcp:60005" in the key


ffourcot:
Applied, thanks!

PR #743, opened 2020-10-26
nftables: Allow negative hook priority and fix typo

blechschmidt:
Base chain priority can be negative (see https://wiki.nftables.org/wiki-nftables/index.php/Configuring_chains). However, since an unsigned value is used, an exception will be thrown when the struct module attempts to pack a negative value that is passed to `NFTables.rule` as the priority parameter. Thus, ff0f00cb309464fa2b406a70ec05abd42f27c0a8 adds signed big endian values.

0c78cd03991786465fe0b01b32a325acea9bf613 fixes a typo.

### DIFF
--- a/pyroute2/netlink/__init__.py
+++ b/pyroute2/netlink/__init__.py
@@ -1696,6 +1696,34 @@ class nlmsg_atoms(nlmsg_base):
 
         fields = [('value', '>Q')]
 
+    class sbe8(nla_base):
+
+        __slots__ = ()
+        sql_type = 'INTEGER'
+
+        fields = [('value', '>b')]
+
+    class sbe16(nla_base):
+
+        __slots__ = ()
+        sql_type = 'INTEGER'
+
+        fields = [('value', '>h')]
+
+    class sbe32(nla_base):
+
+        __slots__ = ()
+        sql_type = 'BIGINT'
+
+        fields = [('value', '>i')]
+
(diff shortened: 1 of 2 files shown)


ffourcot:
Hello,

Interesting. Did you check the kernel code? it looks like policy are not up to date as well:

```net/netfilter/nf_tables_api.c:  [NFTA_HOOK_PRIORITY]    = { .type = NLA_U32 },```

That should be NLA_S32 (I don't think that it has impact on running code, but we are often looking at the policy to convert attributes).

And comment should be fixed in header files:

```include/uapi/linux/netfilter/nf_tables.h: * @NFTA_HOOK_PRIORITY: netfilter hook priority (NLA_U32)```

blechschmidt:
Hi,

I actually checked [include/uapi/linux/netfilter_ipv4.h](https://elixir.bootlin.com/linux/v5.10-rc1/source/include/uapi/linux/netfilter_ipv4.h#L30) and some other places where the hook priority was dealt with as signed integer (such as [/include/linux/netfilter.h](https://elixir.bootlin.com/linux/latest/source/include/linux/netfilter.h#L88)) but did not go in more depth.

ffourcot:
Applied, thanks. If you don't have any objection, I will try to cook a patch to improve variable declaration and documentation in kernel.

blechschmidt:
No objections from my side.

PR #754, opened 2020-12-06
support preferred_lft and valid_lft. fixes #752

### LINKED ISSUE #752: How to set preferred_lft when adding an address?
There does not appear to be a way to set the lifetime of an IPv6 address, using `preferred_lft`. 

Here's what I'm trying to do with pyroute2:
```
with IPRoute() as ipr:
    ipr.link('add', ifname="dummy0", kind='dummy')
    idx = ipr.link_lookup(ifname="dummy0")[0]
    ipr.addr("add", index=idx, address="2001:db8::5678", mask=128, preferred_lft=0)
```

And the result (as we can see `preferred_lft` is not set):

```
$ ip addr show dev dummy0
303: dummy0: <BROADCAST,NOARP> mtu 1500 qdisc noop state DOWN group default qlen 1000
    link/ether d2:37:57:9a:66:e1 brd ff:ff:ff:ff:ff:ff
    inet6 2001:db8::5678/128 scope global
       valid_lft forever preferred_lft forever.              <<<<  preferred lifetime not set 
```

Here's a working example using iproute2 directly, and we can see that the preferred lifetime is set as expected: 

```
$ ip link add dev dummy0 type dummy
$ ip addr add 2001:db8::5678/128 dev dummy0 preferred_lft 0

$ ip addr show dev dummy0
304: dummy0: <BROADCAST,NOARP> mtu 1500 qdisc noop state DOWN group default qlen 1000
    link/ether aa:03:fd:ae:3b:38 brd ff:ff:ff:ff:ff:ff
    inet6 2001:db8::5678/128 scope global deprecated
       valid_lft forever preferred_lft 0sec                <<<<  preferred lifetime set as intended 
```

fritzlab:
In this PR, we support valid lifetime and preferred lifetime for IPv6 addresses.  This fixes #752. 

The following example is now valid: 
```
from pyroute2 import IPRoute
with IPRoute() as ipr:
    ipr.link('add', ifname="dummy0", kind='dummy')
    idx = ipr.link_lookup(ifname="dummy0")[0]
    ipr.addr("add", index=idx, address="2001:db8::5678", mask=128, preferred=123, valid=456)
```

The above example yields the following result: 
```
7340: dummy0: <BROADCAST,NOARP> mtu 1500 qdisc noop state DOWN group default qlen 1000
    link/ether 96:07:42:04:8a:15 brd ff:ff:ff:ff:ff:ff
    inet6 2001:db8::5678/128 scope global dynamic
       valid_lft 456sec preferred_lft 123sec
```

Both `preferred` and `valid` are new kwargs.  when unset the default value of "forever" is used.  Forever is defined by 2^32-1. 

This PR does not add any tests.  The testing suite seems quite complex and, in general, is broken on my machine. It fails when running `make test` on the head of master today.  Even adding a new test function that does nothing fails when running the single test as well.  I'm happy to update this PR if I can be given some guidance on how to get the tests working (ubuntu 20.04, python 3.8.5).

### DIFF
--- a/pyroute2/netlink/__init__.py
+++ b/pyroute2/netlink/__init__.py
@@ -1064,6 +1064,22 @@ class nlmsg_base(dict):
             for name, fmt in self.fields:
                 value = self[name]
 
+                if name == "ifa_valid":
+                    for a, v in self.parent.get("attrs", list()):
+                        if a == "IFA_VALID":
+                            value = v
+                            break
+                    else:
+                        value = pow(2, 32) - 1
+
+                if name == "ifa_preferred":
+                    for a, v in self.parent.get("attrs", list()):
+                        if a == "IFA_PREFERRED":
+                            value = v
+                            break
+                    else:
+                        value = pow(2, 32) - 1
+
                 if fmt == 's':
                     length = len(value)
                     efmt = '%is' % (length)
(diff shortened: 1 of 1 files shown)


svinota:
Thanks, merging

svinota:
But in a separate branch first, as it should not chage the basic netlink decoder (that has nothing to do with ip addresses)

svinota:
Should work now. Thanks for the PR, it helped to fix.

PR #755, opened 2020-12-07
fixed missing tcp diag info symbols

joh2511:
There were some symbols from `struct tcp_info` missing (c.f. `include/uapi/linux/tcp.h`). Those were added now.


svinota:
thanks

PR #758, opened 2020-12-08
Add End.BPF to seg6local

kyontan:
I want to use End.BPF seg6local function with pyroute2.

I saw modified version of https://github.com/Zashas/pyroute2/commit/6fa211a87375dc31bfacedbcb676047a66833c02.  It is partially merged in master of pyroute2, but partially not.

To use End.BPF with `fd` and `name`, I changed some code by refering modified version.

Here is use case of `End.BPF`:

```python3
from bcc import BPF
from pyroute2 import IPRoute

b = BPF(src_file="source.c")
fn = b.load_func("pass", 19)
# 19 means BPF_PROG_TYPE_LWT_SEG6LOCAL 
# https://github.com/torvalds/linux/blob/3cb12d27ff655e57e8efe3486dca2a22f4e30578/include/uapi/linux/bpf.h#L190

ipr = IPRoute()
idx = ipr.link_lookup(ifname=iface)[0]
sid = 'fd00::1/128'

encap = {'type':'seg6local', 'action':'End.BPF', 'bpf':{'fd':fn.fd, 'name':fn.name}}
ipr.route("add", dst=sid, oif=idx, encap=encap)
```

source.c:
```source.c
#include <linux/bpf.h>

int pass(struct __sk_buff *skb) {
    return BPF_OK; // packet continues
}
```

### DIFF
--- a/pyroute2/netlink/rtnl/req.py
+++ b/pyroute2/netlink/rtnl/req.py
@@ -220,6 +220,8 @@ class IPRouteRequest(IPRequest):
             srh = {}
             segs = []
             hmac = None
+            prog_fd = None
+            prog_name = None
             # Parse segs
             if srh:
                 segs = header['srh']['segs']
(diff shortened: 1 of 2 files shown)


svinota:
Cool, thanks

PR #759, opened 2020-12-15
WIP : Extended ack support

ffourcot:
Hello Peter,

During some debug with kernel netlink messages, it appears to us that extended ack support could be a great feature in pyroute2, allowing to read more useful error messages.

I wrote in this MR a proof of concept of how it's working, that one can reproduce with :

```python
In [1]: from pyroute2 import IPRoute

In [2]: ipr = IPRoute(ext_ack=True)

In [3]: ipr.link("dump", if_netnsid=4)  # here 4 must be an invalid netns number
{'error': -22, 'attrs': [('NLMSGERR_ATTR_MSG', 'Invalid target network namespace id')], 'header': {'length': 60, 'type': 3, 'flags': 514, 'sequence_number': 255, 'pid': 30621}}
```

However, there are still a lot of work before to be able to merge it. I didn't find a good way to parse standard nlmsg as nlmsgerr when NLM_F_ACK_TLVS flah is set.

And there are a lot of missing feature, like capped messages, cookies, etc. But I have currently no idea of the best idea to go forward. Could you have a look on it?

### DIFF
--- a/pyroute2/netlink/nlsocket.py
+++ b/pyroute2/netlink/nlsocket.py
@@ -102,13 +102,16 @@ from pyroute2.config import AF_NETLINK
 from pyroute2.common import AddrPool
 from pyroute2.common import DEFAULT_RCVBUF
 from pyroute2.netlink import nlmsg
+from pyroute2.netlink import nlmsgerr
 from pyroute2.netlink import mtypes
 from pyroute2.netlink import NLMSG_ERROR
 from pyroute2.netlink import NLMSG_DONE
 from pyroute2.netlink import NETLINK_ADD_MEMBERSHIP
 from pyroute2.netlink import NETLINK_DROP_MEMBERSHIP
+from pyroute2.netlink import NETLINK_EXT_ACK
 from pyroute2.netlink import NETLINK_GENERIC
 from pyroute2.netlink import NETLINK_LISTEN_ALL_NSID
+from pyroute2.netlink import NLM_F_ACK_TLVS
 from pyroute2.netlink import NLM_F_DUMP
 from pyroute2.netlink import NLM_F_MULTI
 from pyroute2.netlink import NLM_F_REQUEST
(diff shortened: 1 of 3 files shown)


svinota:
Great! Thanks a lot, looking

ffourcot:
Hello @svinota, any news on this topic? I didn't had the time to investigate until now, but I still think that this feature could be great in pyroute2. Any advice to move forward?

Thanks

svinota:
Didn't have time either, but I believe the project **must** have this feature.

Let me play a bit over this weekend and I'll merge it in this or other way.

pyroute2/netlink/nlsocket.py
```diff
@@ -175,6 +178,11 @@ def parse(self, data, seq=None, callback=None):
 
             try:
                 msg.decode()
+                # TODO: not the good method to do that
```
svinota, line 181: Why don't you think it's a proper place to parse an error message?

svinota:
The change looks quite simple, I'm for it

pyroute2/netlink/nlsocket.py
ffourcot: In my opinion, this should be in msg.decode() itself, if possible. It looks weird for me to call msg.decode() twice

And it's not only about error parsing. I hardcoded the nlmsgerr class here, but we can receive some "cookies" with relevant data when extack is enabled

svinota:
Looks like I don't understand something, but can not get it working.

@ffourcot , could you please provide a testcase that would run on net-next 5.11+?

ffourcot:
Hello @svinota,

Yes! I will take some times to provide two scripts (one returning one error, and one with cookies)

svinota:
Got it working, thus I can proceed now with implementation.

Regarding extra decoding — I'll look if it may be done more properly.

svinota:
@ffourcot I plan to raise an exception on `nlmsgerr()` as it's done on ordinary errors.

Any objections / comments ?

ffourcot:
> Got it working, thus I can proceed now with implementation.

Great! Thank you

> @ffourcot I plan to raise an exception on nlmsgerr() as it's done on ordinary errors.

It's perfect from my point of view. The real message from kernel code will really help

svinota:
Merged manually.

The condition to raise the exception is `NLMSG_DONE` + `NLM_F_ACK_TLVS` + `msg['error']`, since`NLM_F_ACK_TLVS` may happen in other dumps as well but w/o any error attached.

PR #776, opened 2021-03-29
[NDB] Change warning to work with Python < 3.6

tmdzk:
Signed-off-by: Timothée Dzik <timdzik@fb.com>

### Summary

In the latest 0.5.15 [this line](https://github.com/svinota/pyroute2/blame/master/pyroute2/ndb/main.py#L307) has been included which work on `Python >= 3.6` but not on `Python < 3.6`

Simple fix, change the way to inject the var in the warning.

### DIFF
--- a/pyroute2/ndb/main.py
+++ b/pyroute2/ndb/main.py
@@ -304,7 +304,7 @@ class Transaction(object):
                     try:
                         rb.rollback()
                     except Exception as e:
-                        self.log.warning(f'ignore rollback exception: {e}')
+                        self.log.warning('ignore rollback exception: %s' % e)
                 raise
         self.event.set()
         return self


tmdzk:
Forgot to introduce myself, I'm Tim currently working on [magma](https://github.com/magma/magma) an open-source software platform that gives network operators an open, flexible and extendable mobile core network solution. 

Thanks for your pyroute2 lib! 👯

svinota:
@tmdzk sure, thanks!

What do you think, may it be a good idea to tag a release?

tmdzk:
hey @svinota, thanks for your fast reply!  I think it is a good idea to tag a release.

svinota:
done, and I hope nothing gets broken :)

PR #777, opened 2021-04-01
ss2: align license with overall libary project

cherusk:
Moving to GPL v2 character of licensing on request
of main maintainer.


cherusk:
It's failing in some VRF test case. That's probably for the testing framework rework you mentioned.

svinota:
Yep. Nevermind the travis testing, it's not fully ready yet thus failing.

All the local tests are passing.

PR #780, opened 2021-04-12
Add support for netem rate setting

mhebant:
Add support for netem rate settings:  rate, packet_overhead, cell_size and cell_overhead


svinota:
Thanks, merging

PR #784, opened 2021-04-21
Fixed "python" in for make README.md rule

### LINKED ISSUE #783: Make for README.md doesn't work on systems without interpreter bin named 'python' (no specific version)
Got this error when trying to run `make install`

```
# make install python=/usr/bin/python3
/bin/sh: python: command not found
make: *** [Makefile:175: README.md] Error 127
```

I think this is due to a typo in the make file using `python` instead of `${python}`.

I have a PR fix incoming.

pgagne:
Fixed #783

When trying to install this on systems the don't have unqualified 'python' (ie no version specified) executable you get an error:

```
# make install python=/usr/bin/python3
/bin/sh: python: command not found
make: *** [Makefile:175: README.md] Error 127
```

My guess is this was a typo and it was intended to use '${python}' for this rule instead, so I fixed it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -172,7 +172,7 @@ dist: force-version docs
 	@${python} setup.py sdist >/dev/null 2>&1
 
 README.md:
-	@cat README.rst | python ./docs/conv.py >README.md
+	@cat README.rst | ${python} ./docs/conv.py >README.md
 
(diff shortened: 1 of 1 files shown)


svinota:
Thanks!

PR #794, opened 2021-05-12
nl80211: parse wiphy bands

### LINKED ISSUE #555: Parsing NL80211_ATTR_WIPHY_BANDS
Can you give me any hints about this?  All the example code I can find online defers it to the libnl parser, but your library doesn't use that parser.  How would I go about parsing this attribute?  Or even where could I find any sort of documentation on its format?

I've tried:

```
bands = filter(lambda x: x[0] == 'NL80211_ATTR_WIPHY_BANDS', data['attrs'])[0]
bands = bytearray(int(x, 16) for x in bands.split(':')
bands = iw.marshal.parse(bands)
```

But this produces something which doesn't make a lot of sense (parsing another layer of nested attributes results in a parse error).

matshch:
This fixes #555 to some extent. I haven't seen `NL80211_BAND_ATTR_IFTYPE_DATA` in the wild, but other information looks fine. Frequencies are easily accessible, and I think most people really interested only in this one. Maybe we should also parse capabilities and present them as individual flags, but I'm not quite sure how to do this in a manner that will fit everybody, so I left them as uints for now.

I don't have much experience with Python's struct format strings, so I left C structure definitions for review.

### DIFF
--- a/pyroute2.core/pr2modules/netlink/nl80211/__init__.py
+++ b/pyroute2.core/pr2modules/netlink/nl80211/__init__.py
@@ -232,7 +232,7 @@ class nl80211cmd(genlmsg):
                ('NL80211_ATTR_STA_SUPPORTED_RATES', 'hex'),
                ('NL80211_ATTR_STA_VLAN', 'hex'),
                ('NL80211_ATTR_STA_INFO', 'STAInfo'),
-               ('NL80211_ATTR_WIPHY_BANDS', 'hex'),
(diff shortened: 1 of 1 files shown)


svinota:
Cool, thanks!

The code will be merged a bit later in the 0.6.0.

matshch:
Rebased commits so they can be merged without problems onto new release

svinota:
Merging, thanks

PR #799, opened 2021-05-19
nl80211: allow to get and set regulatory rules

### LINKED ISSUE #550: Set kernel WiFi regulatory domain
Does pyroute2 provide a way to set the kernel's WiFi regulatory domain?

It looks to me like, if I was clever, I could construct a netlink packet using `NL80211_CMD_REQ_SET_REG` but wondering if there's an easier way before I go digging too deep.

matshch:
Fixes #550.


svinota:
Great, thanks!

PR #800, opened 2021-05-19
nl80211: add new attributes

matshch:
Some attributes in list_wiphy was showing as unknown, so lets add everything that new kernel supports.


svinota:
Thanks!

PR #804, opened 2021-05-24
wireguard: fix parsing keys

### LINKED ISSUE #803: WireGuard keys are parsed incorrectly
Just getting interface info using a simple program like

```python
from pyroute2 import WireGuard
print(WireGuard().info('wg0'))
```

Results in lots of errors like: 

```
decoding WGPEER_A_PUBLIC_KEY
  File "/nix/store/xhifdnnxdif4nrhi434bads4dmb41ik2-python3.9-pyroute2-core-0.6.1/lib/python3.9/site-packages/pr2modules/netlink/__init__.py", line 1507, in try_to_decode
    cell.decode()
  File "/nix/store/xhifdnnxdif4nrhi434bads4dmb41ik2-python3.9-pyroute2-core-0.6.1/lib/python3.9/site-packages/pr2modules/netlink/generic/wireguard.py", line 158, in decode
    self['value'] = b64encode(self['value'])
  File "/nix/store/xhifdnnxdif4nrhi434bads4dmb41ik2-python3.9-pyroute2-core-0.6.1/lib/python3.9/site-packages/pr2modules/netlink/__init__.py", line 1194, in __getitem__
    return dict.__getitem__(self, key)
KeyError: 'value'
```

And similar for `WGDEVICE_A_PRIVATE_KEY`, `WGDEVICE_A_PRESHARED_KEY` and `WGPEER_A_PRESHARED_KEY`.

A device object is still returned, but the keys are one byte long: `'WGDEVICE_A_PRIVATE_KEY', {'key': b'\xc8'}`.

Using Python 3.9 and pyroute2 0.6.1 on NixOS.

K900:
Load the keys from the correct place in the message, and load them as the right length as well. Fixes #803.

### DIFF
--- a/pyroute2.core/pr2modules/netlink/generic/wireguard.py
+++ b/pyroute2.core/pr2modules/netlink/generic/wireguard.py
@@ -151,11 +151,11 @@ class wgmsg(genlmsg):
                    ('WGPEER_A_PROTOCOL_VERSION', 'uint32'))
 
         class parse_peer_key(nla):
-            fields = (('key', 's'), )
+            fields = (('key', '32s'), )
 
             def decode(self):
                 nla.decode(self)
-                self['value'] = b64encode(self['value'])
+                self['value'] = b64encode(self['key'])
 
             def encode(self):
                 self['key'] = b64decode(self['value'])
@@ -220,11 +220,11 @@ class wgmsg(genlmsg):
                 self['addr'] = '{0}/{1}'.format(self['addr'], wgaddr)
 
     class parse_wg_key(nla):
-        fields = (('key', 's'), )
+        fields = (('key', '32s'), )
 
         def decode(self):
             nla.decode(self)
-            self['value'] = b64encode(self['value'])
+            self['value'] = b64encode(self['key'])
 
         def encode(self):
             self['key'] = b64decode(self['value'])


pyroute2.core/pr2modules/netlink/generic/wireguard.py
```diff
@@ -151,11 +151,11 @@ class wgdevice_peer(nla):
                    ('WGPEER_A_PROTOCOL_VERSION', 'uint32'))
 
         class parse_peer_key(nla):
-            fields = (('key', 's'), )
+            fields = (('key', '32s'), )
 
             def decode(self):
                 nla.decode(self)
-                self['value'] = b64encode(self['value'])
+                self['value'] = b64encode(self['key'])
```
K900, line 158: We might want to also `.decode()` this, so that the keys are strings, the same way they're accepted in other places.

svinota: Yep, I'll take a look.

svinota:
Thanks!

pyroute2.core/pr2modules/netlink/generic/wireguard.py
K900: Please tell me what you decide on, because that's going to determine how I should handle this on my application's side.

Issue #826, opened 2021-07-30
get_netns_info doesn't work if using with NetNS


antonkurbatov:
I ran into a problem when get_netns_info returns unknown (0xffffffff) value for netnsid.

Here is a simple demo:

Create a veth pair and set NS for each end
```
$ ip link add veth1 type veth peer name veth2
$ ip netns add ns1
$ ip netns add ns2
$ ip link set veth1 netns ns1
$ ip link set veth2 netns ns2
```

So, `ip a` for ns1 looks like:
```
$ ip netns exec ns1 ip a
1: lo: <LOOPBACK> mtu 65536 qdisc noop state DOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
43: veth1@if42: <BROADCAST,MULTICAST> mtu 1500 qdisc noop state DOWN group default qlen 1000
    link/ether 36:e0:7c:58:17:79 brd ff:ff:ff:ff:ff:ff link-netns ns2
$
```

Then use 'get_netns_info' to get the `netnsid`, it returns unknown (0xffffffff) value:

```
$ python3
>>> import pyroute2
>>> with pyroute2.NetNS('ns1', flags=0) as ip:
...     for item in ip.get_netns_info():
...         if item.get_attr('NSINFO_PATH').endswith('ns2'):
...             print('%s -> %s' % (item.get_attr('NSINFO_PATH'), item['netnsid']))
...
/var/run/netns/ns2 -> 4294967295
>>>
```

At the same time, if enter into ns1 namespace:
```
$ ip netns exec ns1 bash
$ python3
>>> import pyroute2
>>> ip = pyroute2.IPRoute()
>>> for item in ip.get_netns_info():
...     if item.get_attr('NSINFO_PATH').endswith('ns2'):
...         print('%s -> %s' % (item.get_attr('NSINFO_PATH'), item['netnsid']))
...
/var/run/netns/ns2 -> 1
>>>
```

It looks like the main reason is how get_netns_info works: it opens /var/ run/netns/ns2 after the NetNS fork & setns, and the child process doesn't know any about the opened FD:

```
$ python3
>>> import pyroute2
>>> import os
>>> from pr2modules.netlink.rtnl.nsidmsg import nsidmsg
>>> from pr2modules.netlink.rtnl import RTM_GETNSID
>>> from pr2modules.netlink import NLM_F_REQUEST
>>>
[...]

PR #838, opened 2021-08-27
[core, nslink] Open/close files in child process

centromere:
File descriptors are not shared between parent and child after `fork()`.
Therefore, calls to `os.open()`/`os.close()` must be proxied.

See #826.


svinota:
Thanks!

PR #839, opened 2021-08-27
[ndb] Add is_peer() method for interface objects

centromere:
This determines whether a given interface "points to" another. It is relevant to vlans and veths.

This also fixes #835.

PR #842, opened 2021-09-10
[core, ndb, nslink] Use different approach for is_peer()

centromere:
Addresses the issue discussed in #839.


svinota:
Thanks, merging as is, later I'll change to support V[X]LAN as well

PR #848, opened 2021-09-28
Fix compat with Sphinx 4.x

thomasgoirand:
Pyroute2 conf.py is still using app.add_stylesheet() which was deprecated in
the favor of app.add_css_file(). In 4.x, app.add_stylesheet() was completely
removed, rendering Pyroute2's doc incompatible with Sphinx 4.x This patch
fixes it.


svinota:
Thanks!

PR #872, opened 2022-01-10
Fix occasional crashes in NSPopen.release

### LINKED ISSUE #871: NSPopen occasionally crashes in release() with "Bad File Descriptor"
Occasionally the release() function for an NSPopen object crashes with an OSError (Bad File Descriptor) exception.

```
  File "y.py", line 10, in <module>
    ns_exec.release()
  File "/opt/amphora-agent-venv/lib/python3.8/site-packages/pr2modules/nslink/nspopen.py", line 343, in release
    self.channel_out._writer.close()
  File "/usr/lib/python3.8/multiprocessing/connection.py", line 177, in close
    self._close()
  File "/usr/lib/python3.8/multiprocessing/connection.py", line 361, in _close
    _close(self._handle)
OSError: [Errno 9] Bad file descriptor
```

I can reproduce that on an Ubuntu 20.04 VM with python 3.8.10, pyroute2 0.6.4 and the following script:

```
import pyroute2
import subprocess

i = 0
while True:
    i += 1
    print(i)
    ns_exec = pyroute2.NSPopen("some-namespace", ["lsmod"], stdout=subprocess.PIPE)
    ns_exec.wait()
    ns_exec.release()
```

Sometimes the loop continues for 100 cycles, but often the exception occurs already after less then 10 cycles.

This seems to be related to a patch made to fix an FD leak (issue #622, PR #779).
Without that patch (e.g. with pyroute2 0.5.14) the multiprocessing.Queue FD leak will hit and the test loop above crashes after some 338 cycles with "Too many open files".

I found corresponding bug tickets for the multiprocessing module (https://bugs.python.org/issue33081 and https://bugs.python.org/issue42752). At least for 33081 the ticket wos closed with "won't fix", so I understand that NSPopen needs to implement some workaround. I tried with a different approach than explicitly closing `self.channel_out._writer`. It may be enough to delete `self.channel_out` which will then trigger the Queue cleanup and let it close the writer FD. For me the following patch to NSPopen solved the problems:

```
diff --git a/pyroute2.nslink/pr2modules/nslink/nspopen.py b/pyroute2.nslink/pr2modules/nslink/nspopen.py
index 7430912e..91436881 100644
--- a/pyroute2.nslink/pr2modules/nslink/nspopen.py
+++ b/pyroute2.nslink/pr2modules/nslink/nspopen.py
@@ -339,9 +339,9 @@ class NSPopen(ObjNS):
             self.channel_in.close()
             self.server.join()
             # clean leftover pipes that would be closed at program exit
-            os.close(self.server.sentinel)
-            self.channel_out._writer.close()
-            self.channel_in._writer.close()
+            del self.server
+            del self.channel_out
+            del self.channel_in

     def __dir__(self):
         return list(self.api.keys()) + ['release']
```

bpetermannS11:
Fix for the issue that NSPopen's release() occasionally crashes
with an OSError / "Bad File Descriptor" inside
self.channel_out._writer.close().
To still avoid the file descriptor leak in multiprocessing.Queue
but be more robust: Switch from closing the writer explicity to
deleting the Queue object.

Fixes #871

### DIFF
--- a/pyroute2.nslink/pr2modules/nslink/nspopen.py
+++ b/pyroute2.nslink/pr2modules/nslink/nspopen.py
@@ -8,7 +8,6 @@ namespace support.
 
 '''
 
-import os
 import sys
 import fcntl
 import types
@@ -339,9 +338,9 @@ class NSPopen(ObjNS):
             self.channel_in.close()
             self.server.join()
             # clean leftover pipes that would be closed at program exit
-            os.close(self.server.sentinel)
-            self.channel_out._writer.close()
-            self.channel_in._writer.close()
+            del self.server
+            del self.channel_out
+            del self.channel_in
 
     def __dir__(self):
         return list(self.api.keys()) + ['release']


svinota:
Thanks!

PR #887, opened 2022-03-15
Fixed unnecessary reloading of libc

hazimeh:
When setns (and other libc functions) are called repeatedly, they load libc at every function call, which results in a lot of I/O overhead.

This fix allows the loaded libc module to be cached globally and fetched later when required.

### DIFF
--- a/pyroute2.core/pr2modules/netns/__init__.py
+++ b/pyroute2.core/pr2modules/netns/__init__.py
@@ -118,7 +118,7 @@ MS_SHARED = 1 << 20
 NETNS_RUN_DIR = '/var/run/netns'
 
 __saved_ns = []
-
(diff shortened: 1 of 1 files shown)


svinota:
True. Thanks, merging!

ffourcot:
@svinota usage of ":=" operator will reduce number of python supported version.

It's fine for me, but it's the first usage of this operator in pyroute2 code base.

svinota:
Sure thing, I run tests from 3.6 (see `tox.ini`). In this case it was easier to merge and fix :) Waiting for the CI to finish and pushing the fix.

And @ffourcot  — thanks **a lot** for keeping en eye on the code.

svinota:
updated

PR #889, opened 2022-03-20
IPDB: Interface Address AF_INET6

svenauhagen:
On Point to Point connections like PPPoE the local ip address
is in IFA_LOCAL and not in IFA_ADDRESS.
IFA_ADDRESS is the remote ip address in that case.
Check for IFA_LOCAL first before using IFA_ADDRESS.

### DIFF
--- a/pyroute2.ipdb/pr2modules/ipdb/interfaces.py
+++ b/pyroute2.ipdb/pr2modules/ipdb/interfaces.py
@@ -1407,7 +1407,9 @@ class AddressesDict(dict):
         if msg['family'] == AF_INET:
             addr = msg.get_attr('IFA_LOCAL')
         elif msg['family'] == AF_INET6:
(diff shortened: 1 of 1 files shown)


svinota:
Sure thing I merge that. But keep in mind that IPDB is deprecated.

If you can not replace IPDB code with NDB (it's not drop-in, some adaptation is required) — pls report missing functionality, it will be added asap.

svenauhagen:
Thanks. Last time I tried NDB it was pretty slow querying the SQLite in memory DB. I have not looked into it. Probably because it is single read/write when in memory.
I will give it another try soon.

svinota:
Got it. I will double check the performance bottlenecks in NDB.

svenauhagen:
Maybe you can create an option to enable WAL mode when using the SQLite DB as a file or in general to set some options after the DB is created. That would allow multiple read and single write in parallel. It would benefit large setups.

PR #899, opened 2022-04-22
iproute/linux: try to improve flags when sending del messages

ffourcot:
Flags are not the same between NEW and DELETE request. In file netlink.h
we can see:

 /* Modifiers to NEW request */
 #define NLM_F_REPLACE   0x100   /* Override existing        */
 #define NLM_F_EXCL      0x200   /* Do not touch, if it exists   */
 #define NLM_F_CREATE    0x400   /* Create, if it does not exist */
 #define NLM_F_APPEND    0x800   /* Add to end of list       */

 /* Modifiers to DELETE request */
 #define NLM_F_NONREC    0x100   /* Do not delete recursively    */
 #define NLM_F_BULK      0x200   /* Delete multiple objects  */

So NLM_F_BULK is the same value than NLM_F_EXCL. NLM_F_BULK has been
added in this kernel commit:

https://git.kernel.org/pub/scm/linux/kernel/git/netdev/net-next.git/commit/?id=a6cec0bcd34264be8887791594be793b3f12719f

When this flag is set, pyroute now get a lot of error everywhere when
trying to delete objects.

Note that kernel commit is not yet mainline, and still in development
branch. But it can have a huge impact on pyroute2 library

### DIFF
--- a/pyroute2.core/pr2modules/iproute/linux.py
+++ b/pyroute2.core/pr2modules/iproute/linux.py
@@ -665,7 +665,7 @@ class RTNL_API(object):
             # flush all addresses with IFA_LABEL='eth0':
             ipr.flush_addr(label='eth0')
         '''
-        flags = NLM_F_CREATE | NLM_F_EXCL | NLM_F_REQUEST
+        flags = NLM_F_CREATE | NLM_F_REQUEST
         ret = []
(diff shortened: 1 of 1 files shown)


svinota:
Awesome. Thanks a lot!

PR #902, opened 2022-04-23
nftables: Add more operations and raise kernel errors

blechschmidt:
Hello,

this PR is motivated by the lack of error reporting for nftables write operations. I have seen that there are unfinished classes like `BatchSocket` and `BatchAddrPool`, which are probably supposed to tackle the issue which I am addressing here. So I hope that this PR is not too intrusive or diverging too much from your intention.

Currently, the handling of batch messages is difficult, as `nlm_request` sends each message separately and listens for responses from a single message sequence number only. Therefore, 39741a4660d25cccf59f750dbb354d8c0fc16e0c introduces `nlm_request_batch`, which can be supplied with a list of messages. The function will then listen for expected responses for any of the supplied messages and return/yield them to the caller.

This is leveraged by 729de2d55d998e47af773dc4cfb385ddd9841d75, which introduces acknowledgements for write operations by default. With this commit, errors, e.g. caused by invalid rule expressions or missing NLAs, will no longer go unnoticed and silent failures will no longer happen. The commit also adds some additional commands for the `table`, `chain` and `rule` objects.

Before, the following would silently fail because the chain does not exist:
`nft.rule('add', table='test', chain='doesnotexist', expressions=some_valid_expr)`

Now it will fail as indicated by the errno from the kernel:
`pr2modules.netlink.exceptions.NetlinkError: (2, 'No such file or directory')`

Silent failures can still be enabled by setting the `flags` parameter to 0. If the parameter is not set, it will default to `NLM_F_ACK`, which will cause the batch request function to expect an acknowledgement:
`nft.rule('add', table='test', chain='doesnotexist', expressions=some_expr, flags=0)`

Kind regards.

### DIFF
--- a/pyroute2.core/pr2modules/common.py
+++ b/pyroute2.core/pr2modules/common.py
@@ -477,6 +477,26 @@ class AddrPool(object):
             else:
                 raise KeyError('no free address available')
 
+    def alloc_multi(self, count):
+        with self.lock:
+            addresses = []
+            raised = False
+            try:
+                for _ in range(count):
+                    addr = self.alloc()
+                    try:
+                        addresses.append(addr)
+                    except:  # In case of a MemoryError during appending, the finally block would not free the address.
+                        self.free(addr)
+                return addresses
+            except:
+                raised = True
+                raise
+            finally:
+                if raised:
+                    for addr in addresses:
+                        self.free(addr)
+
     def locate(self, addr):
         if self.reverse:
             addr = self.maxaddr - addr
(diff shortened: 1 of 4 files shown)


svinota:
* formatting issues — I have to update the contribution docs, thanks for the reminder
* nlm_request_batch — there was a plan to introduce something like `BatchSocket.commit()` also

Let's merge your code first: It's better to have something working and used, and thereafter we can eventually employ other possible solutions, if there will be any.

Thanks for the contribution!

PR #924, opened 2022-06-09
iptun: fix IFLA_IPTUN_FLAGS types

zeil:
https://elixir.bootlin.com/linux/v5.18.3/source/net/ipv6/ip6_tunnel.c#L2191


svinota:
Thanks for the fix!

PR #925, opened 2022-06-09
add explicit dependency ordering in setup target

foxxx0:
Both the `dist` and `uninstall` targets had listed `clean` and `VERSION`
listed as dependencies *before* `setup`.
However, these dependencies need to be performed in this particular
order and `Makefile` dependencies do not guarantee dependency ordering
when executed with more than one job.

As the `clean` and `VERSION` targets are only needed for `setup`, we can
move their execution to be always part of `setup` and then enforce
ordering by explicitly calling them sequentially.

Some discussion on this can be found here:
https://stackoverflow.com/questions/8496135/parallel-makefile-requires-dependency-ordering/8496739#8496739

Additionally this commit replaces the `ls` parsing with `find` and moves
templating listing into a separate method to deduplicate the code.
This obsoletes the need for doing `sed` on the output and generally
parsing the output of `ls` is not a great idea:
https://mywiki.wooledge.org/ParsingLs

If we were to only target Linux coreutils, we could make use of the
`-printf '%f\n'` feature, but the BSD variants lack support for
`-printf`, which is why the `-print0` + `xargs -0 basename` was
selected.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,7 +48,11 @@ endif
 # Functions
 #
 define list_modules
-	`ls -1 | sed -n '/egg-info/n; /pyroute2/p'`
+	`find . -maxdepth 1 -mindepth 1 -type d -iname 'pyroute2*' -print0 | xargs -0 -n1 basename`
+endef
+
+define list_templates
+	`find templates -maxdepth 1 -mindepth 1 -type f -print0 | xargs -0 -n1 basename`
 endef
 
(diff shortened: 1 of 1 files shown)


svinota:
CI error in functional tests is in a flaky test, that's already fixed in the master, I will take care of it.

svinota:
Looks reasonable. Thanks!

PR #961, opened 2022-07-18
setup: specify "build-backend" in pyproject.toml

henning-schild:
Fixes: 9714df3404b5 ("setup: remove setup.py")

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,2 +1,3 @@
 [build-system]
 requires = ["setuptools", "wheel"]
+build-backend = "setuptools.build_meta"


svinota:
@henning-schild thanks for the fix!

Do you think that tagging a new release would help also?

henning-schild:
> @henning-schild thanks for the fix!
> 
> Do you think that tagging a new release would help also?

In gentoo i am currently proposing a package with the fix applied, i do not mind having a cherry-pick in that. Others might prefer a patch-free or might ask me to mark it as a cherry and name the upstream commit sha.

For gentoo i guess we do not need a new one. Others might appreciate it, trying to package but failing ... just to find that master carries a fix.

PR #965, opened 2022-07-19
Fix default action in RuleIPRouteFilter

### LINKED ISSUE #964: rule add creates invalid rule entries
Using pyroute2 0.7.1, adding rules leads to incorrect entries in the rule list, and the rules are not correctly applied by the kernel.

(Issue detected in the OpenStack Octavia project)

I reproduced it with this simple script:

```
import pyroute2

with pyroute2.IPRoute() as ipr:
    ipr.rule('add', src="10.1.1.1", src_len=32, table=1)
```

then ip rule shows a strange entry that contains a none field:

```
$ ip rule | grep 10.1
32000:  from 10.1.1.1 lookup 1 none
```

Note: 0.6.13 doesn't have this issue

gthiemonge:
During the refactoring of the ip rule module, the default action when
adding a rule with a table was incorrectly re-written. Adding a rule
with a table parameter created incorrect rule entries.

Fixes #964

### DIFF
--- a/pyroute2/requests/rule.py
+++ b/pyroute2/requests/rule.py
@@ -28,7 +28,7 @@ class RuleIPRouteFilter(IPRouteFilter):
                 context['family'] = socket.AF_INET
             if 'priority' not in context:
                 context['priority'] = DEFAULT_FRA_PRIORITY
-            if 'table' not in context and 'action' not in context:
+            if 'table' in context and 'action' not in context:
                 context['action'] = 'to_tbl'
             for key in ('src_len', 'dst_len'):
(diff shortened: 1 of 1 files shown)


elajkat (commented):
We found in Neutron this one also, thanks for it, see: https://bugs.launchpad.net/neutron/+bug/1981963

A workaround of it in Neutron: https://review.opendev.org/c/openstack/neutron/+/850223

svinota:
Got it, thanks a lot! I have no idea why I put such a weird negation into the rule processing, thanks for fixing it.

0.7.2 with this change will be tagged tonight.

Issue #989, opened 2022-08-10
protocol state mismatch between conntrack and pyroute2


ohashemzadeh:
Hello.
There is difference between pyroute2 and Conntrack output (mismatch)
In below code, when we get state from **conntrack output** , it returns **TIME_WAIT**
**but** pyroute2 Conntrack, returns two different states **STABLISHED,SYN_SENT**.

```python3
from typing import List, Generator, Final
from pr2modules.conntrack import Conntrack
from pr2modules.conntrack import ConntrackEntry
from pr2modules.netlink.nfnetlink.nfctsocket import NFCTAttrTuple

conntrack = Conntrack()
conntrack_dumped_entries: Generator = conntrack.dump_entries(tuple_orig=NFCTAttrTuple())
conntrack_entries: List[ConntrackEntry] = list(conntrack_dumped_entries)

# Extracting my record (named `my_conntrack_entry`) ....
# my_conntrack_entry : ConntrackEntry = my_method_to_extract_my_record()
print (str(my_conntrack_entry.protoinfo.state_name))
```

Here is pyroute2 output:
```
"<bound method NFCTATcpProtoInfo.state_name of TcpInfo(state=ESTABLISHED,SYN_SENT, orig_flags=WINDOW_SCALE,SACK_PERM,CLOSE_INIT,MAXACK_SET, reply_flags=WINDOW_SCALE,SACK_PERM,MAXACK_SET)>"
```

conntrack -L output:
```
tcp      6 117 TIME_WAIT src=192.168.200.200 dst=192.168.115.15 sport=48654 dport=22 src=192.168.115.15 dst=192.168.200.200 sport=22 dport=48654 [ASSURED] mark=0 use=1
```

```
Using pyroute2 version 0.7.2
on Ubuntu 20.04
conntrack v1.4.5
```

svinota:
Thanks, investigating

svinota:
@ohashemzadeh pls check if we can close the issue

PR #991, opened 2022-08-11
Conntrack tcp state fix

inemajo:
Hello, this fix the issue #989 (conntracks use it own tcp status with a enum and not flags)


svinota:
@inemajo could you please run `make format` and update the PR? It would significantly simplify the merge :)

Thanks a lot for the fix!

svinota:
Thanks!

Issue #995, opened 2022-08-13
NDB listening on specific netlink groups


cunha:
We use `pyroute2` on software routers with millions of routes. It seems NDB loads a full snapshot of the namespace in its database, which in our case takes a significant time.

Is there a way to control what `netlink` groups NDB binds to?

If not, we would be willing to contribute changes to allow choosing which `netlink` groups NDB listens on. I am thinking of a similar change to the one we contributed back in 2016 to support the same functionality on IPDB: https://github.com/svinota/pyroute2/pull/314

Looking at the code, I thought of two ways to achieve this:

1. Allow users to pass initialized `IPRoute` or `NetNS` as NDB sources. The user would be free to bind these `IPRoute` and `NetNS` instances to specific `netlink` groups as desired. (As of right now, it seems that only dictionaries specifying sources are accepted; `IPRoute` and `NetNS` are instantiated afterwards.) This is more-or-less what the user interface would look like:

    ```{python}
    netns = NetNS('mynamespace')
    netns.bind(groups=desired_netlink_groups)
    src = { 'kind': 'netns', 'instance': netns }
    with NDB(sources=[src]) as ndb:
        ...
    ```

2. Allow users to pass desired `netlink` groups as part of the NDB source specification. Then, the groups could be passed to `netlink` when [binding the socket](https://github.com/svinota/pyroute2/blob/fddc1b35e4cbbb878264f50eab77095a44762c68/pyroute2/ndb/source.py#L352). This is what the user interface would look like:

    ```{python}
    desired_netlink_groups = RTMGRP_DEFAULTS & (~RTMGRP_IPV4_ROUTE) & (~RTMGRP_IPV6_ROUTE)
    src = { 'netns': 'mynamespace', 'netlink_groups': desired_netlink_groups }
    with NDB(sources=[src]) as ndb:
        ...
    ```

Any preference or feedback on any of these approaches? (It seems to me 2 is easier to use and easier to implement!)

Would listening on a subset of groups violate NDB's assumptions? For example, if one configures NDB to not listen to route messages (like in the example above), then we could understandably raise an exception if the user accesses `ndb.routes`, but it's unclear to me whether NDB requires listening to all groups (in particular, it would need to support maintaining a partial database).

svinota:
The reason why NDB accepts a parameters dict instead of an instance is that NDB tries to restart sources in the case of failures. 

[...]

PR #997, opened 2022-08-16
iproute, netns: support netlink_groups

svinota:
Bug-Url: https://github.com/svinota/pyroute2/issues/995

### DIFF
--- a/pyroute2/iproute/linux.py
+++ b/pyroute2/iproute/linux.py
@@ -259,7 +259,7 @@ class RTNL_API:
 
     # 8<---------------------------------------------------------------
     #
-    def dump(self, groups=0xFFFFFFFF):
+    def dump(self, groups=None):
         '''
         Dump network objects.
 
@@ -311,7 +311,7 @@ class RTNL_API:
                 RTMGRP_IPV6_RULE: [partial(self.get_rules, family=AF_INET6)],
             }
         for group, methods in groups_map.items():
-            if group & groups:
+            if group & (groups if groups is not None else self.netlink_groups):
                 for method in methods:
                     for msg in method():
                         yield msg
--- a/pyroute2/iproute/linux.py
+++ b/pyroute2/iproute/linux.py
@@ -60,6 +60,14 @@ from pyroute2.netlink.rtnl import (
     RTM_NEWTCLASS,
     RTM_NEWTFILTER,
     RTM_SETLINK,
+    RTMGRP_IPV4_IFADDR,
+    RTMGRP_IPV4_ROUTE,
+    RTMGRP_IPV4_RULE,
+    RTMGRP_IPV6_IFADDR,
+    RTMGRP_IPV6_ROUTE,
+    RTMGRP_IPV6_RULE,
+    RTMGRP_LINK,
+    RTMGRP_NEIGH,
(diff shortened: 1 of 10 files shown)


pyroute2/iproute/linux.py
```diff
@@ -60,6 +60,14 @@
     RTM_NEWTCLASS,
     RTM_NEWTFILTER,
     RTM_SETLINK,
+    RTMGRP_IPV4_IFADDR,
+    RTMGRP_IPV4_ROUTE,
+    RTMGRP_IPV4_RULE,
+    RTMGRP_IPV6_IFADDR,
+    RTMGRP_IPV6_ROUTE,
+    RTMGRP_IPV6_RULE,
```
ffourcot, line 68: I do not found RTMGRP_IPV6_RULE constant in current kernel headers. Is it a typo with RTNLGRP_IPV6_RULE ? (RTM vs RTN)

ffourcot: Ok, it looks like all RTMGRP_* variables are outdated and working only as compat.

See https://lwn.net/Articles/147608/ for reference

svinota: There is already `NetlinkSocket.add_membership()/drop_membership()`, so probably I have to change NDB sources to use it. Let's schedule it as a separate task.

cunha:
I went through the changes and they make sense to me. I also ran some local tests and can confirm that the new `netlink_groups` option significant improves NDB creation time. It covers my use-case successfully.

One thing to consider/document in the API is that we can pass in `netlink_groups` as a parameter when creating `IPRoute` and `NetNS` instances. The `netlink_groups` are later passed on to `IPRoute.bind()` and `NetNS.bind()`. However, `netlink_groups` are not passed on to `dump()`, which instead receives a `groups` parameter.

I can contribute changes to the docs to cover this parameter and its use around early September.

cunha:
What I was getting at is that `dump()` does not inherit the `netlink_groups` passed to `__init__()`, while `bind()` does. I am not saying this is a problem, I just wanted to mention it for awareness.

I think we can leave everything as-is, including `bind(groups)` to avoid the deprecation. We just need to update the configuration to match the code (which I can do in a few weeks).

svinota:
Yep, I tried now to change `groups` to `netlink_groups`, but it affects too much unrelated code.

`dump()` does not inherit `netlink_groups` 'cause simple dump commands (`link('dump')`, `route('dump')` etc) are not affected by the `bind()` arguments, they return the same results.

But you correctly point this out, it may be an issue, since it's another inconsistency.

svinota:
Finally dropping `netlink_groups` in favour of `groups`. The reasons:

* constency with existing `bind()`
* the attribute is to be used by PF_ROUTE sockets as well, and PF_ROUTE ≠ netlink, thus `netlink_groups` is not the best naming

Also, made `dump()` to respect `self.groups` by default, if not forced by the method argument.

svinota:
@cunha , @ffourcot — if no objections, I would merge.

cunha:
Ran my (rather minimal) tests again without issues.

ffourcot:
@svinota it looks great, thanks!

Issue #1013, opened 2022-08-31
How to add ip address into nftables's sets?


rexzhang:
I have pre-created the table and sets.
```
>>> from pyroute2.nftables.main import NFTables
>>> nft=NFTables(nfgen_family=7)
>>> nft.get_sets()
[...]

PR #1017, opened 2022-09-05
Draft: Nftables sets

inemajo:
Hello @svinota, this is a draft for the issue #1013 (implementation of sets and set_elements for Nftables)

### DIFF
--- a/pyroute2/netlink/nfnetlink/nftsocket.py
+++ b/pyroute2/netlink/nfnetlink/nftsocket.py
@@ -16,6 +16,7 @@ from pyroute2.netlink import (
     NLM_F_REPLACE,
     NLM_F_REQUEST,
     nla,
+    nlmsg_atoms,
 )
 from pyroute2.netlink.nfnetlink import NFNL_SUBSYS_NFTABLES, nfgen_msg
 from pyroute2.netlink.nlsocket import NetlinkSocket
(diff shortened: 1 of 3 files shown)


pyroute2/netlink/nfnetlink/nftsocket.py
```diff
@@ -46,6 +47,55 @@
 NFT_MSG_GETFLOWTABLE = 23
 NFT_MSG_DELFLOWTABLE = 24
 
+# from nftables/include/datatype.h
```
svinota, line 50: Just a comment, what do you think — would we benefit of using `enum` in these cases instead of a plain definition of variables?

Not starting with this PR, but in some future maybe.

pyroute2/nftables/main.py
```diff
@@ -125,3 +137,83 @@ def rule(self, cmd, **kwarg):
...
+        if modifier is not None:
+            elements = []
+            for elem in kwarg.pop("elements"):
+                modifier.value = elem
+                modifier.encode()
+                elements.append(modifier["value"])
+        else:
+            kwarg.pop("elements")
+
+        kwarg["elements"] = [
```
svinota, line 211: I already started to work on kwarg filters for `IPRoute`, see `/pyroute2/requests/`; eventually we could make some common API that solves common transformations like `str 255.255.255.0 ⇒ int 24`, or `iterable ( int interface_index ) ⇒ int interface_index`

svinota:
LGTM, if you could fix pep8 (run `make format`) — would be nice.

I see no issues so far, but tonigh I'll look more closer and start some test routines as well.

pyroute2/netlink/nfnetlink/nftsocket.py
inemajo: Yes I think it would be better

pyroute2/nftables/main.py
inemajo: I will see that thanks !

svinota:
@inemajo does the functional test succeeds on your system?

inemajo:
Yes, and I try on a VM with ubuntu-18 and they pass too, so I don't understand why they don't pass on the CI

inemajo:
I put some new features on sets (timeout, counter...) I want to add userdata (for the comment entry) and after passing the MR to ready

inemajo:
that's what I thought, the test create the filter table but it had no effect.

Thanks for the help

inemajo:
Hello @svinota, I move the tests into examples for unlock the CI, do you think it's ok for you ?

svinota:
Sure thing, thanks.

Issue #813, opened 2021-06-03
DEFAULT_RCVBUF too small when querying interfaces with more than 108 VFs


gibizer:
Hi, 

Similar to https://github.com/svinota/pyroute2/issues/751 if the number of VFs are increased over 108 then the buffer in pyroute2 becomes too small.

> <180>2021-05-31T16:35:00.707793+02:00 compute-0-4 neutron-sriov-nic-agent[33389]: 2021-05-31 16:35:00.706 455 WARNING pyroute2.netlink [-] Traceback (most recent call last):
>   File "/usr/lib/python3.6/site-packages/pyroute2/netlink/__init__.py", line 1345, in _ft_decode_generic
>     self.decode_nlas(offset)
>   File "/usr/lib/python3.6/site-packages/pyroute2/netlink/__init__.py", line 1470, in decode_nlas
>     offset)
> struct.error: unpack_from requires a buffer of at least 4 bytes
> <179>2021-05-31T16:35:00.731008+02:00 compute-0-4 neutron-sriov-nic-agent[33389]: 2021-05-31 16:35:00.707 6 ERROR neutron.plugins.ml2.drivers.mech_sriov.agent.sriov_nic_agent [req-34c81c8f-7a7a-4a75-8ce9-c2b91cfdba21
>] Error in agent loop. Devices info: {}: TypeError: can not serialize 'error' object
> 2021-05-31 16:35:00.707 6 ERROR neutron.plugins.ml2.drivers.mech_sriov.agent.sriov_nic_agent Traceback (most recent call last):
> 2021-05-31 16:35:00.707 6 ERROR neutron.plugins.ml2.drivers.mech_sriov.agent.sriov_nic_agent   File "/usr/lib/python3.6/site-packages/neutron/plugins/ml2/drivers/mech_sriov/agent/sriov_nic_agent.py", line 472, in daemon_loop
> 2021-05-31 16:35:00.707 6 ERROR neutron.plugins.ml2.drivers.mech_sriov.agent.sriov_nic_agent     device_info = self.scan_devices(devices, updated_devices_copy)
> 2021-05-31 16:35:00.707 6 ERROR neutron.plugins.ml2.drivers.mech_sriov.agent.sriov_nic_agent   File "/usr/lib/python3.6/site-packages/osprofiler/profiler.py", line 160, in wrapper
[...]

Issue #751, opened 2020-11-24
DEFAULT_RCVBUF  to small when querying interfaces with large number of VF's


javacruft:
I have a deployment where an SR-IOV card has a large number of VF's configured - 63.

When pyroute2 (0.5.9) is used to interrogate the interface I see the following error:

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3/dist-packages/pyroute2/iproute/linux.py", line 1330, in link
    ret = self.nlm_request(msg,
  File "/usr/lib/python3/dist-packages/pyroute2/netlink/nlsocket.py", line 373, in nlm_request
    return tuple(self._genlm_request(*argv, **kwarg))
  File "/usr/lib/python3/dist-packages/pyroute2/netlink/nlsocket.py", line 862, in nlm_request
    for msg in self.get(msg_seq=msg_seq,
  File "/usr/lib/python3/dist-packages/pyroute2/netlink/nlsocket.py", line 376, in get
    return tuple(self._genlm_get(*argv, **kwarg))
  File "/usr/lib/python3/dist-packages/pyroute2/netlink/nlsocket.py", line 701, in get
    raise msg['header']['error']
  File "/usr/lib/python3/dist-packages/pyroute2/netlink/nlsocket.py", line 177, in parse
    msg.decode()
  File "/usr/lib/python3/dist-packages/pyroute2/netlink/rtnl/ifinfmsg/__init__.py", line 1087, in decode
    nlmsg.decode(self)
  File "/usr/lib/python3/dist-packages/pyroute2/netlink/__init__.py", line 1315, in _ft_decode_generic
    raise NetlinkNLADecodeError(e)
pyroute2.netlink.exceptions.NetlinkNLADecodeError: unpack_from requires a buffer of at least 19240 bytes for unpacking 4 bytes at offset 19236 (actual buffer size is 16384)
```

Doubling the size of the RCVBUF resolves the issue and appears to be valid for later kernel versions where the maximum message size has increased to 32K

svinota:
Got it, thanks! Will be fixed.

PR #1045, opened 2022-10-18
Add _peek_bufsize to NetlinkSocketBase so recv can use it

### LINKED ISSUE #1044: DEFAULT_RCVBUF can be (again) small with more than 127 VFs
In our lab we had NICs with more and more VFs with recent increase of DEFAULT_RCVBUF  to 65536, we can have more than 127 VFs. The previous attempts to push out this limit (https://github.com/svinota/pyroute2/issues/813 or https://github.com/svinota/pyroute2/issues/751 ) just pushed out the issue but not solved it.

One of our experts  checked the problem, and suggested a solution to detect the real buffer size for receiving netlink, something like this in NetlinkSocketBase:
from socket import MSG_DONTWAIT, MSG_PEEK, MSG_TRUNC
def peek_bufsize(socket_descriptor):
    data = bytearray()
    bufsize, _ = socket_descriptor.recvfrom_into(data, 0, MSG_DONTWAIT | MSG_PEEK | MSG_TRUNC)
    return bufsize

def recv(self, *argv, **kwarg):
    ....
+    return self._sock.recv(self._peek_bufsize(self._sock), **kwarg)
-     return self._sock.recv(*argv, **kwarg)

This can be easily check in a dummy environment (I used a VM without sriov or any extra HW):
import sys
from pyroute2 import IPRoute

device_idx = int(sys.argv[1])
with IPRoute() as ip:
    print(ip.link('get',
                  index=device_idx,
                  ext_mask=1))

And changing DEFAULT_RCVBUF to 1024.
I got NLMSG_ERROR with the original code, and with peek_bufsize I just got the listing of links as I expected.

elajkat:
NetlinkSocketBase.recv uses DEFAULT_RCVBUF to fetch things like ip links via netlink sockets. It happed in the last cycles that with new NICs with higher VF numbers or similar we have to change the buffer size (see #751 or #813).
A better solution (thanks to one or our experts) can be to check the necessary buffer size before receiving the sockets.

Fixes #1044

### DIFF
--- a/pyroute2/netlink/nlsocket.py
+++ b/pyroute2/netlink/nlsocket.py
@@ -92,7 +92,15 @@ import time
 import traceback
 import warnings
 from functools import partial
-from socket import MSG_PEEK, SO_RCVBUF, SO_SNDBUF, SOCK_DGRAM, SOL_SOCKET
+from socket import (
+    MSG_DONTWAIT,
+    MSG_PEEK,
+    MSG_TRUNC,
+    SO_RCVBUF,
+    SO_SNDBUF,
+    SOCK_DGRAM,
+    SOL_SOCKET,
+)
 
 from pyroute2 import config
 from pyroute2.common import DEFAULT_RCVBUF, AddrPool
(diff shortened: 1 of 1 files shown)


svinota:
Great, thanks!

PR #1048, opened 2022-10-25
nlsocket: fix BlockingIOError on socket peek ops

svinota:
@elajkat pls take a look -- unconditional peek causes BlockingIOError on wireless ops.

I tried to fix, and I hope it doesn't break your cases.


elajkat:
Thanks for notifying, looks ok, and as I checked works for our scenario

svinota:
Thanks!

PR #1052, opened 2022-11-15
Added NDA_SRC_VNI to support vlan-aware vxlan tunnels.

MatthewKazmar:
Added NDA_SRC_VNI to support vlan-aware vxlan tunnels.


svinota:
Looks like I have an issue with one of the CI test workers, fixing it now.

PR #1056, opened 2022-11-28
fix vid/vni range

MatthewKazmar:
Fix an error with VLAN and VNI ranges.


svinota:
Thanks!

Btw, Mattew, any plans to cover with tests? Could I help with that?

MatthewKazmar:
Thanks, Peter.

I haven't written tests before. Let me look at the tests you have written. I will try to add cases to cover my changes.
________________________________

PR #1057, opened 2022-12-06
dhcp/__init__.py: fix bytes vs. str in literal

keithkostol:
Looks like this string literal didn't get changed over to be a byte string during the python3 transition.


svinota:
Indeed. Thanks!

PR #1062, opened 2023-01-03
install ss2 as module as well

cherusk:
prometheus_ss_exporter depends on the logic also as module.

Installing it as cli and tool makes it accessible conventiently to both ways of consumption.


cherusk:
@svinota would be great to get a minor release for that please.

cherusk:
https://github.com/cherusk/prometheus_ss_exporter/pull/6

svinota:
Will be ready on Monday. Sorry for delay.

Issue #1079, opened 2023-03-14
ethtool file descriptor leakage


socketpair:
```python
#!/usr/bin/python3

from pyroute2.ethtool import Ethtool  # pylint: disable=no-name-in-module,import-error
import os
import subprocess
import gc

def main():
    subprocess.check_call(['ls', '-l', f'/proc/{os.getpid()}/fd'])
    Ethtool()
    gc.collect()
    subprocess.check_call(['ls', '-l', f'/proc/{os.getpid()}/fd'])

if __name__ == '__main__':
    main()
```

shows:

```
# ./zzz.py 
итого 0
lrwx------ 1 root root 64 мар 14 19:06 0 -> /dev/pts/3
lrwx------ 1 root root 64 мар 14 19:06 1 -> /dev/pts/3
l-wx------ 1 root root 64 мар 14 19:06 11 -> 'pipe:[1220330]'
lrwx------ 1 root root 64 мар 14 19:06 2 -> /dev/pts/3
l-wx------ 1 root root 64 мар 14 19:06 9 -> 'pipe:[1220329]'
итого 0
lrwx------ 1 root root 64 мар 14 19:06 0 -> /dev/pts/3
lrwx------ 1 root root 64 мар 14 19:06 1 -> /dev/pts/3
l-wx------ 1 root root 64 мар 14 19:06 11 -> 'pipe:[1220330]'
lrwx------ 1 root root 64 мар 14 19:06 2 -> /dev/pts/3
lr-x------ 1 root root 64 мар 14 19:06 4 -> 'pipe:[1237864]'
l-wx------ 1 root root 64 мар 14 19:06 5 -> 'pipe:[1237864]'
l-wx------ 1 root root 64 мар 14 19:06 9 -> 'pipe:[1220329]'
```

as you can see, three pipes leaked.

socketpair:
```
    subprocess.check_call(['ls', '-l', f'/proc/{os.getpid()}/fd'])

    NlEthtool()
    gc.collect()

    subprocess.check_call(['ls', '-l', f'/proc/{os.getpid()}/fd'])
```

the same

socketpair:
```python
    subprocess.check_call(['ls', '-l', f'/proc/{os.getpid()}/fd'])

    GenericNetlinkSocket()
    gc.collect()

    subprocess.check_call(['ls', '-l', f'/proc/{os.getpid()}/fd'])
```
the same

socketpair:
```
[...]

PR #1081, opened 2023-03-15
ethtool fd leaks

svinota:
Bug-Url: https://github.com/svinota/pyroute2/issues/1079

### DIFF
--- a/pyroute2/ethtool/ethtool.py
+++ b/pyroute2/ethtool/ethtool.py
@@ -257,12 +257,18 @@ class EthtoolLinkMode(
         )
 
 
-class Ethtool(object):
+class Ethtool:
     def __init__(self):
         self._with_ioctl = IoctlEthtool()
         self._with_nl = NlEthtool()
         self._with_nl.module_err_level = 'debug'
         self._is_nl_working = self._with_nl.is_nlethtool_in_kernel()
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        self.close()
 
     def _nl_exec(self, f, with_netlink, *args, **kwargs):
         if with_netlink is None:
@@ -345,4 +351,5 @@ class Ethtool(object):
         self._with_ioctl.set_coalesce(ioctl_coalesce)
 
     def close(self):
+        self._with_ioctl.close()
         self._with_nl.close()
--- a/tests/test_linux/test_ethtool.py
+++ b/tests/test_linux/test_ethtool.py
@@ -14,3 +14,11 @@ def test_pipe_leak():
     etht.close()
     gc.collect()
     assert get_fds() == fds
+
+
+def test_context_manager():
+    fds = get_fds()
+    with Ethtool():
+        pass
+    gc.collect()
+    assert get_fds() == fds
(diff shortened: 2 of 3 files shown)


pyroute2/ethtool/ethtool.py
```diff
@@ -257,13 +257,19 @@ def from_netlink(cls, nl_link_mode):
...
 
-class Ethtool(object):
+class Ethtool:
     def __init__(self):
         self._with_ioctl = IoctlEthtool()
         self._with_nl = NlEthtool()
         self._with_nl.module_err_level = 'debug'
         self._is_nl_working = self._with_nl.is_nlethtool_in_kernel()
 
+    def __enter__(self):
```
socketpair, line 267: isn't's it better to define this in one of the basic classes ?

tests/test_linux/test_ethtool.py
```diff
@@ -0,0 +1,28 @@
+import gc
+import os
+
+from pyroute2 import Ethtool
+
+
+def get_fds():
+    fd = os.open(f'/proc/{os.getpid()}/fd', os.O_RDONLY)
+    try:
+        return set(os.listdir(fd))
```
socketpair, line 10: `return set(os.listdir(fd)) - {fd}`

svinota: Indeed.

socketpair:
lgtm

tests/test_linux/test_ethtool.py
```diff
@@ -0,0 +1,28 @@
...
+    try:
+        return set(os.listdir(fd)) - {fd}
+    finally:
+        os.close(fd)
+
+
+def test_pipe_leak():
+    fds = get_fds()
+    etht = Ethtool()
+    etht.close()
```
socketpair, line 18: HUH. not calling `.close()` will leak descriptors anyway ?  You have to add `__del__` to classes that use `os.open()` / `os.pipe()` in class constructors.

svinota: Personally I prefer resources to be freed explicitly, putting anyting in `__del__` requires a lot of checks and exception handling to guarantee no errors during the GC work.

But I'm to try to look what we can do there.

socketpair: @svinota just close the resources opened explicitly. 

```
from contextlib import suppress

class Xxx:
    def __init__(self):
            self.__x, self.__y = os.pipe()
            
     def close():
            # i.e. check every resource if it is closed already, close and clear the variable somehow.
            if self.__x !=1:
                os.close(self.__x)
                self.__x = -1 # not 'None' because of typing.
          
    def __del__(self):
            super().__del__() # is it required?
            self.close()

```

socketpair: Another way:
```python
class FdWrapper(int):
    def __del__(self):
        if not hasattr(self, '__closed'):
            try:
                print('here')
                os.close(self)
            except OSError:
                pass  # TODO: log ?
            finally:
                setattr(self, '__closed', True)


def MyPipe() -> tuple[FdWrapper, FdWrapper]:
    x, y = os.pipe()
    return FdWrapper(x), FdWrapper(y)


x, y = MyPipe()
os.write(y, b'qwe')

```

PR #1086, opened 2023-04-04
nl80211 decoder additions

Vic063:
Here is a set of patches to improve the decoder for nl80211 messages.

It allows to decode:
 - WiPhy supported commands
 - WiPhy cipher suites
 - WiPhy supported interface types (monitor, ap, managed, and so on...)

It also set the correct type to decode the number of TX/RX antennas.


svinota:
Thanks, merged!

PR #1092, opened 2023-05-08
Add support for CGROUP_ID to inet_diag_message and fix type for CLASS_ID

PaulFurtado:
I currently use `inet_diag_msg` to look up the class id of sockets from the application's net_cls cgroup and I've been decoding the hex string, but always meant to make a PR to fix it to be uint32. I'm now upgrading the system to cgroup v2, so I need to move to cgroup id instead which required adding additional fields.

You can see where these are set upstream in the kernel here: https://github.com/torvalds/linux/blob/ac9a78681b921877518763ba0e89202254349d1b/net/ipv4/inet_diag.c#L170-L177
These fields are populated when you request the TCLASS extension.


svinota:
Thanks!

PR #1098, opened 2023-05-26
sched_netem: fix duplicate attribute value for netem qdisc

BrianBaboch:
Just like the loss option, the duplicate option is given as a percentage so we should convert that percentage to a value in before using it.


svinota:
Thanks!

PR #1109, opened 2023-07-07
Introduce netlink/rtnl CAN support

### LINKED ISSUE #1023: setup CAN interfaces
Hi svinota,
I 've some issues concerning CAN interfaces.
With IPRoute module, i try to create CAN interfaces and it's not possible: in my linux, when you plug your CAN device, it was created by the system. You just need set the interface and not create it. 
So when you try to create the CAN interface, you have the message "File exist" and it's normal.
My issue is focus on set the interfaces.
I've got this scenario: 
If I plug my Can device on USB and do nothing, I can't  set "Status" to "UP"
idem for flags parameter: flags=128 is running well (this is down status of the interface) 
flags=193 doesn't modify the interface and set it to UP status.

BUT 

in a bash shell, if I setup my CAN interface with: 
ip link set can1 up type can bitrate 500000

After this command, 
all modifications under all items like flags, status , ... is running well and is effective.

Perhaps, i mess something or I don't follow the right process.

I've trying to do the same things with the same scenario with NDB module and I 've the same thing.
For information: i use the 0.7.2 version of PyRoute2

B.R

johnernberg:
Adds handling of all the attributes related to CAN.

I was not able to figure out how create an interface similar to the iproute package, so the interface is a bit more low level in that instead of adding bitrate with "bitrate=<rate>" one has to package a can_bittiming dict with {"bitrate": <rate>}.

The attribute CAN_CLOCK is a struct of one member in the ABI, but it doesn't seem that fields support one-member structs.

A couple of the attributes haven't been tested because I do not have any dongles that support them. E.g. the TDC options.

Fixes #1023

### DIFF
--- a/pyroute2/netlink/rtnl/ifinfmsg/__init__.py
+++ b/pyroute2/netlink/rtnl/ifinfmsg/__init__.py
@@ -12,6 +12,7 @@ from pyroute2.config import AF_BRIDGE
 from pyroute2.netlink import NLA_F_NESTED, nla, nlmsg, nlmsg_atoms
 from pyroute2.netlink.rtnl.ifinfmsg.plugins import (
     bond,
+    can,
     geneve,
     gtp,
     ipoib,
@@ -257,6 +258,7 @@ data_plugins = {}
 
 for module in (
     bond,
+    can,
     geneve,
     gtp,
     ipvlan,
--- a/pyroute2/netlink/rtnl/ifinfmsg/plugins/can.py
+++ b/pyroute2/netlink/rtnl/ifinfmsg/plugins/can.py
@@ -1,0 +1,154 @@
+from pyroute2.netlink import nla, nlmsg_atoms
+
+CAN_CTRLMODE_NAMES = {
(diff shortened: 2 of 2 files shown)


svinota:
Thanks a lot!

yegorich:
Some info about my system. It's Debian Bookworm (Linux 6.1.0). CAN hardware is an SLCAN device.

The following command is working as expected:

`ip link set can0 type can bitrate 100000`

johnernberg:
@yegorich As explained in the commit message this is a little bit more low level than the iproute command line.

What iproute tools call type is here called 'kind' because that is what netlink uses internally. And the bitrate must be provided using the can_bittiming structure which has a bitrate field.

Try:
```
ip_route.link('set', index=can0, kind='can', can_bittiming={'bitrate': 100000 })
```

yegorich:
@johnernberg it works like a charm. Thank you very much. I'll try to submit a PR with such an example.

@svinota where should I place this example code? `examples/iproute`?

PR #1113, opened 2023-07-21
Allow passing args up to GenericNetlinkSocket constructor

kleinweby:
Currently, most subclasses of GenericNetlinkSocket do not pass unused args up to GenericNetlinkSocket.__init__.

One use case of this is to pass fileno= with an existing file descriptor that was opened in another process/network namespace.

This patch fixes the following subclasses

  - EventSocket
  - NL80211
  - DevlinkSocket
  - L2tp
  - WireGuard

by passing up *args and **kwargs.

We also fix TaskStats by simply removing the __init__ method that just calls super.

We did not need to fix MPTCP and NlEthtool as they don't overwrite __init__.

### DIFF
--- a/pyroute2/netlink/devlink/__init__.py
+++ b/pyroute2/netlink/devlink/__init__.py
@@ -492,8 +492,8 @@ class MarshalDevlink(Marshal):
 
 
 class DevlinkSocket(GenericNetlinkSocket):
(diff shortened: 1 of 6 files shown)


svinota:
Thanks a lot!

PR #1121, opened 2023-08-11
iproute: add support for netns nsid allocation

liske:
I need to assign a nsid to peer netns and it looks like that pyroute2 is currently missing a `RTM_NEWNSID` (`ip netns set`) implementation. However, I am not sure if my approach for adding this request is good.


svinota:
Thanks!

PR #1126, opened 2023-08-24
ethtool: statistics support

pspacek:


pspacek:
@svinota Hello! I don't want to be inpatient, but what do you think about this PR? :-)

svinota:
Sorry for delay, I was having some issues.

And thanks for your patience

PR #1134, opened 2023-10-24
l2tp: Fix get_tunnel/get_session to return only requested object

kleinweby:
L2TP does not support using filters on NLM_F_DUMP and always returns all objects. The current get_tunnel/session functions however always use NLM_F_DUMP and set the tunnel/session ids if provided by the caller. These are simply ignored.

This patch fixes get_tunnel and get_session to always require specifying the desired tunnel/session id to always get one requested object. If the requested object does not exist an NetlinkError is thrown.

To also provide a way to get all tunnels/sessions dump_tunnels and dump_sessions has been introduced.

---

Note that does does change the signature of `get_tunnel` and `get_session`. I've also adjusted the return value to return the one `l2tpmsg` message we get, so users don't need a odd `[0]` anymore.

### DIFF
--- a/pyroute2/netlink/generic/l2tp.py
+++ b/pyroute2/netlink/generic/l2tp.py
@@ -537,35 +537,50 @@ class L2tp(GenericNetlinkSocket):
 
         return self._do_request(msg)
 
-    def get_tunnel(self, tunnel_id=None):
+    def get_tunnel(self, tunnel_id):
         """
-        Get one or more tunnels
-        :param tunnel_id: tunnel id of the tunnel to show, if not set all
-                          tunnels will be returned
+        Get one tunnel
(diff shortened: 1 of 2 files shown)


svinota:
I'm working on fixing the CI pipeline, the PR will be merged this weekend

doeringudh:
I believe there might have happened an oversight with regard to the parameters for the new function dump_tunnels() in pyroute2/netlink/generic/l2tp.py when merging this PR's original patch. Said function has a parameter "tunnel_id" that is actually never used, and also would not make sense in this context.

Now compare it with the also new and related dump_sessions(), which correctly does not have any parameters, apart from "self". In my view dump_tunnels() should have the same signature.

PR #1143, opened 2023-11-17
Add an example for setting CAN bitrate

yegorich:


yegorich:
This example shows new feature inrtoduced in #1109.

PR #1144, opened 2023-11-17
Examples: use python3 consistently

yegorich:


svinota:
TBH, I would rather use `python` everywhere instead of `python3`, but you can convince me.

yegorich:
There is no such thing as `python` anymore. For example, in Debian Bookworm there is only `python3` binary and no symlink as `python`. At least, this is my experience after Python2 was deprecated.

svinota:
Related: https://peps.python.org/pep-0394/

svinota:
Thanks, merged. Will try to keep it consistent.

PR #1147, opened 2023-11-18
Makefile: use nox as a Python module

yegorich:
Debian packages don't provide `nox` file. Hence, use it as a module.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@
 python ?= $(shell util/find_python.sh)
 
 define nox
-	nox $(1) -- '$(subst ",\",${noxconfig})'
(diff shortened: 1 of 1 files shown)


yegorich:
Debian's pytho3-nox filelist:

https://packages.debian.org/sid/all/python3-nox/filelist

Makefile
```diff
@@ -6,7 +6,7 @@
 python ?= $(shell util/find_python.sh)
 
 define nox
-	nox $(1) -- '$(subst ",\",${noxconfig})'
+	python3 -m nox $(1) -- '$(subst ",\",${noxconfig})'
```
svinota, line 9: Why not use python variable from line 6 here?

PR #1152, opened 2023-12-07
ethtool: add ring support to ethtool

BrianBaboch:
Added support to be able to modify and display the ring parameters in ethtool like we would do using `ethool -g` for display and `ethtool -G` for modification

By default we would try to use netlink for the ring operations with a fallback to ioctl in case netlink is not supported:


svinota:
Wow. Thanks!

Issue #1091, opened 2023-05-03
NDB: fdb multicast dump missing records


pkulev:
pyroute2 version: 0.6.13 and 0.7.7

Steps to reproduce:

I'm using a little homebrew cache for multicast records:
```python
In [17]: constants.MCAST_LLADDR
Out[17]: '00:00:00:00:00:00'

In [18]: vtep.fdb.refresh??
Signature: vtep.fdb.refresh() -> 'None'
Docstring: Clear the cache and fill data.
Source:
    def refresh(self) -> None:
        self.clear()

        for fdb in self.root.ip.fdb("dump"):
            if (
                fdb.get_attr("NDA_DST") is not None
                and fdb.get_attr("NDA_LLADDR") == constants.MCAST_LLADDR
            ):
                entry = FDB(
                    fdb["ifindex"],
                    fdb.get_attr("NDA_LLADDR"),
                    fdb.get_attr("NDA_DST"),
                )
                self.fdb.setdefault(entry.index, {})[entry.dst] = entry
Type:      method
```

If we refresh it and show the internals, there will be two records.

```python
In [19]: vtep.fdb.refresh()

In [20]: vtep.fdb.fdb
Out[20]:
{32: {'10.1.0.107': FDB(index=32, lladdr='00:00:00:00:00:00', dst='10.1.0.107'),
  '10.1.0.11': FDB(index=32, lladdr='00:00:00:00:00:00', dst='10.1.0.11')}}
```

If we will do similar filtering using NDB: we will get only one record:
```python
In [21]: fdb_dump = ndb.fdb.dump()

In [22]: fdb_dump.select_records(lladdr=constants.MCAST_LLADDR, dst=lambda dst: dst is not None)

In [23]: fdb_dump.select_fields("lladdr", "dst", "ifindex")

In [24]: fdb_dump
Out[24]: ('00:00:00:00:00:00', '10.1.0.11', 32)
```

bridge fdb dump shows also two records:
```
~# bridge fdb show | grep 00:00:00:00:00:00
00:00:00:00:00:00 dev vxln7 dst 10.1.0.107 self extern_learn permanent
00:00:00:00:00:00 dev vxln7 dst 10.1.0.11 self extern_learn permanent
```

It looks like overriding with lastly added record.
If I manually add a new record via `bridge fdb append` NDB notices this and shows only the last record in the same dump.

How can I help you in investigating that mismatch? Maybe there is workaround?

svinota:
[...]

PR #1158, opened 2023-12-25
ndb: make non-unique FDB index more unique

pkulev:
I found out that brmsg_schema has unique index on (ifindex, lladdr, vlan), so that gives an answer to me what happend in the linked issue.

There can be multicast FDBs with single ifindex and no NDA_VLAN at all (that means NDA_VLAN = 0), but NDA_DST will be different.

We can clearly see that in the example from the issue:
```
~# bridge fdb show | grep 00:00:00:00:00:00
00:00:00:00:00:00 dev vxln7 dst 10.1.0.107 self extern_learn permanent
00:00:00:00:00:00 dev vxln7 dst 10.1.0.11 self extern_learn permanent
```

This change of index breaks backward compatibility for filters like this:
```python
fdb = ndb.fdb.dump()
fdb.select_records(dst=lambda dst: dst is not None)
do_something_with(list(fdb))
```
Now it's more like identity check:
```python
fdb.select_records(dst=lambda dst: dst)
```

*Closes:* #1091

### DIFF
--- a/pyroute2/ndb/objects/neighbour.py
+++ b/pyroute2/ndb/objects/neighbour.py
@@ -48,8 +48,9 @@ ndmsg_schema = (
 
 brmsg_schema = (
     ndmsg.sql_schema()
-    .unique_index('ifindex', 'NDA_LLADDR', 'NDA_VLAN')
+    .unique_index('ifindex', 'NDA_LLADDR', 'NDA_DST', 'NDA_VLAN')
     .constraint('NDA_LLADDR', "NOT NULL DEFAULT ''")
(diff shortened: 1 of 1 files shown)


pkulev:
@svinota could you please review this PR and tell me what do you think about it?

svinota:
Thanks, looking how to merge (a small change required, but I fix it)

svinota:
Pls check if this will work for you.

pkulev:
Thanks! I'll test changes tomorrow.

svinota:
Sure thing, to be tagged tonight

PR #1179, opened 2024-02-17
NDB: Route Replace Error

svenauhagen:
When modifying an existing route with a different
Gateway and Interface the current code does not query the existing route correctly from the database.
Hence we end up with 2 routes in the database instead of replaceing the existing one.

We need to specifically limit the count query to the unique attributes of a route that is replaced which are DST, DST_LEN, Priority and Table.


svinota:
Thanks a lot! Testing

svinota:
Thanks, merged!

PR #1189, opened 2024-03-11
NDB: Address removed netlink message

svenauhagen:
Currently only direct routes are removed from the database if the last interface address is removed.
We also need to delete next hops that are associated with the interface and afterwards remove
routes without next hops.


svinota:
Great, thanks!

PR #1210, opened 2024-09-26
switch euid/ruid field names for uid/gid events

sigint2:
change `proc_event_uid` and `proc_event_gid` to return `ruid/euid` and `rgid/egid` respectively.   currently the values are using the union of each event type.

great feature by the way!


svinota:
Thanks for the contribution!

Issue #1214, opened 2024-10-04
NDB: drop NEWNEIGH netlink event if several FDB replace occured


cactusFriday:
Hello! I've run into some issue with drop NEWNEIGH netlink events.

## What is going on
I'm running latest master branch of pyrotue2, Python 3.9.19
I have instantiated NDB instance on application setup:
```python
class Manager:
    def __init__(self) -> None:
        self.ip = IPRoute()
        self.ndb = NDB(log="debug" if os.getenv("DEBUG") else False)
```
OS:
```
[root@dev23 ~]# uname -a
Linux dev23 4.18.0-477.27.1.el8_8.x86_64 #1 SMP Wed Sep 20 15:55:39 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux
```
I have some records in FDB, but it's not important
```
[root@dev23 ~]# bridge fdb show
...
6e:ce:aa:48:5e:ca dev br-vxlan vlan 1 master br-vxlan permanent
6e:ce:aa:48:5e:ca dev vxln18 vlan 28 master br-vxlan permanent
6e:ce:aa:48:5e:ca dev vxln18 master br-vxlan permanent
00:00:00:00:00:00 dev vxln18 dst 172.31.16.22 self extern_learn permanent
d0:fe:00:00:00:22 dev vxln18 dst 172.31.16.22 self extern_learn sticky
0a:01:c5:06:3a:a1 dev vxln18 dst 172.31.16.22 self extern_learn sticky
```

## What i do to get a `drop`
I just do several similar FDB replaces (doesn't matter whether it's done via `bridge fdb replace` or `self.ip.fdb("replace")`)
Let's evaluate:
```
[root@dev23 ~]# bridge fdb replace 6a:5e:0d:03:33:25 dev vxln18 dst 10.1.0.35 self extern_learn sticky

$$ in bridge monitor fdb
6a:5e:0d:03:33:25 dev vxln18 dst 10.1.0.35 self extern_learn sticky permanent
6a:5e:0d:03:33:25 dev vxln18 self permanent

$$ nothing in pyroute logs (all events handled)

$$ second replace
[root@dev23 ~]# bridge fdb replace 6a:5e:0d:03:33:25 dev vxln18 dst 10.1.0.35 self extern_learn sticky

$$ in bridge monitor fdb
6a:5e:0d:03:33:25 dev vxln18 self permanent

$$ pyroute starts to rescheduling events as can't handle it
[...]

PR #1215, opened 2024-10-04
reorder af_bridge_fdb table indexes in a right way

cactusFriday:
issue: https://github.com/svinota/pyroute2/issues/1214


svinota:
Merging, thanks. It looks like I did a mistake defining the index.

PR #1218, opened 2024-11-08
Adding kwarg parameter to ipmock

Gollam:
Pylint is reporting a false positive unexpected-keyword-arg. At first I thought it was a pylint bug, so I reported it: https://github.com/pylint-dev/pylint/issues/9922. But as it turned out, the bug is in the ipmock module inside pyroute2. Some functions lack a **kwarg parameter, so it needs to be added there, that's why I created this PR.


svinota:
Thanks!

PR #1222, opened 2024-12-18
Conntrack status filtering

ffourcot:
Hello,

A small pull request adding ability to filter conntrack by status / status_mask.

This is interesting since it allows to filter for invalid connections, for example


svinota:
Thanks a lot!

I'm in the middle of migrating to the new core, and will rebase the PR asap.

svinota:
@ffourcot pls notice that now we have asincio under the hood.

The plan is to keep the sync API compatible as much as we can, but now the async API is the first class citizen.

Please let me know if there will be issues with the new core (and I'm pretty sure there will be)

ffourcot:
@svinota thank you! We will test intensively. Our production is a little late from master/recent pyroute2 version, but async API is a very nice feature

PR #1229, opened 2025-01-03
fix state and noarp flag setting in LinkFieldFilter

vmsh0:
I submit this pull request which fixes setting the state and noarp flags in LinkFieldFilter. Previously, the flags were not combined together due probably due to an operator precedence oversight.

First, a commit introduces two tests:

- A trivial `state='up'` test which passes
- A combined `state='up'` and `noarp=True` test which fails due to LinkFieldFilter failing to combine the two flags

Then, a further commit fixes the underlying issue:
- Parenthesization fixes the operator precedence issue
- The hardcoded value `1` is replaced with constant `IFF_UP`


svinota:
Thanks, merged! A special thank you for creating the tests for this PR!

Issue #1232, opened 2025-01-06
Regression on IPRoute.tc


ffourcot:
Hello,

tc() function is not working as before, it looks like arguments changed somewhere:

```python
In [4]: IPRoute().tc('add', 'sfq', 4)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In [4], line 1
----> 1 IPRoute().tc('add', 'sfq', 4)

File pyroute2/pyroute2/iproute/linux.py:2481, in IPRoute.__getattr__.<locals>.synchronize_generic(*argv, **kwarg)
   2479 else:
   2480     task = collect_op
-> 2481 return self.event_loop.run_until_complete(task())

File /usr/lib/python3.11/asyncio/base_events.py:653, in BaseEventLoop.run_until_complete(self, future)
    650 if not future.done():
    651     raise RuntimeError('Event loop stopped before Future completed.')
--> 653 return future.result()

File pyroute2/pyroute2/iproute/linux.py:2471, in IPRoute.__getattr__.<locals>.synchronize_generic.<locals>.collect_op()
   2470 async def collect_op():
-> 2471     return await symbol(*argv, **kwarg)

TypeError: RTNL_API.tc() takes from 2 to 3 positional arguments but 4 were given

In [5]: IPRoute().tc('replace', 'ingress', 3)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In [5], line 1
----> 1 IPRoute().tc('replace', 'ingress', 3)

File pyroute2/pyroute2/iproute/linux.py:2481, in IPRoute.__getattr__.<locals>.synchronize_generic(*argv, **kwarg)
   2479 else:
   2480     task = collect_op
-> 2481 return self.event_loop.run_until_complete(task())

File /usr/lib/python3.11/asyncio/base_events.py:653, in BaseEventLoop.run_until_complete(self, future)
    650 if not future.done():
    651     raise RuntimeError('Event loop stopped before Future completed.')
--> 653 return future.result()

File pyroute2/pyroute2/iproute/linux.py:2471, in IPRoute.__getattr__.<locals>.synchronize_generic.<locals>.collect_op()
   2470 async def collect_op():
-> 2471     return await symbol(*argv, **kwarg)

TypeError: RTNL_API.tc() takes from 2 to 3 positional arguments but 4 were given
```

svinota:
Should be fixed now.

ffourcot:
Thank you!

PR #1233, opened 2025-01-07
iproute: support compat tc args

svinota:
Bug-Url: https://github.com/svinota/pyroute2/issues/1232

### DIFF
--- a/pyroute2/iproute/linux.py
+++ b/pyroute2/iproute/linux.py
@@ -1771,7 +1771,7 @@ class RTNL_API:
             return request.response()
         return [x async for x in request.response()]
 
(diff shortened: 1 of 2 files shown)


pyroute2/iproute/linux.py
```diff
@@ -1872,6 +1872,13 @@ async def tc(self, command, kind=None, **kwarg):
             if kind is None:
                 raise ValueError('must specify kind for non-dump commands')
             kwarg['kind'] = kind
+        # 8<-----------------------------------------------
+        # compatibility section, to be cleaned up?
```
ffourcot, line 1876: Note that documentation (line 1809 for example) is still using "old" syntax

PR #1236, opened 2025-01-07
Draft: requests/tc: do not try to fix del requests

ffourcot:
Hello,

This is a proof of concept of a fixup for del-class command. It looks like "fix_request" and other magic for tc are breaking a simple command:

```python
ipr = IPRoute()
ipr.tc('del-class', 'htb', if_index, classid)
```

We do not need TCA_HTB_PARMS argument for this command (and perhaps some other? I did not check).


svinota:
Let me test it a bit.

ffourcot:
See https://github.com/svinota/pyroute2/pull/1239

PR #1250, opened 2025-01-14
WIP: add an async dhcp client

### LINKED ISSUE #1243: DHCP client
Hello,

I'm currently working on an async DHCP client for our needs. Florent pointed me to the existing DHCP code in pyroute2, it looks like a great base to start from. Would you be okay with me "taking over" this part of the code in a PR ?

Here's my current plan:
- Switch the existing raw / DHCP socket implementation to asyncio (i got this working locally)
- Write an async client with states and hooks, using the aforementioned DHCP socket for communication (got a basic POC working already)

I'm trying to adapt & write some tests before I push anything. I saw you used udhcpd for this purpose previously, would you mind me using dnsmasq in tests ?

Thanks,
Étienne

etene:
Here is my async DHCP client implementation, still a work in progress

closes #1243


pyroute2/dhcp/constants/dhcp.py
```diff
@@ -0,0 +1,130 @@
...
+    INTERNET_STORAGE_NAME_SERVICE = 83  # ISNS
+    NDS_SERVERS = 85  # Novell Directory Services Servers
+    NDS_TREE_NAME = 86  # Novell Directory Services Tree Name
+    NDS_CONTEXT = 87  # Novell Directory Services Context
+    BCMCS_CONTROLLER_DOMAIN = 88  # BCMCS Controller Domain Name List
+    AUTHENTICATION = 90  # Authentication
+    CLIENT_SYSTEM_ARCHITECTURE_TYPE = 93  # Client System Architecture Type
+    CLIENT_NETWORK_INTERFACE_IDENTIFIER = 94  # Client Network Interface Identifier
+    CLASSLESS_STATIC_ROUTE = 121
+    DOMAIN_SEARCH = 119
```
ffourcot, line 128: this one is not sorted?

svinota:
Wow, c'est magnifique. It will take some time to read & digest, be patient :)

etene:
> Wow, c'est magnifique. It will take some time to read & digest, be patient :)

Thanks ! Take your time, it's still a draft and lots of things are going to change, i'm not yet proud with everything

pyroute2/ext/rawsocket.py
```diff
@@ -55,28 +55,37 @@ class RawSocket(socket):
...
                     break
             else:
                 raise IOError(2, 'Link not found')
-        self.l2addr = link.get_attr('IFLA_ADDRESS')
-        self.ifindex = link['index']
+        self.l2addr: str = link.get_attr('IFLA_ADDRESS')
+        self.ifindex: int = link['index']
         # bring up the socket
         socket.__init__(self, AF_PACKET, SOCK_RAW, htons(ETH_P_ALL))
+        socket.setblocking(self, False)
```
svinota, line 73: @etene what would you say — could it be a reasonable idea to make a socket wrapper that would use composition, and proxy most of the properties, but fake `.type` to make it work with `loop.create_datagram_endpoint()` ? and then use `asyncio.Protocol`

etene: I think it's an interesting idea.

I have tried at some point to make raw sockets work with `asyncio.Protocol` but changed my mind when i saw how a lot of code assumed TCP, UDP or Unix sockets iirc. But I hadn't thought about faking the socket type, and it might just work given how similar raw & UDP sockets are.

I'm willing to try if I find the time to. However, the python stdlib code might be too tightly coupled with UDP/TCP/unix sockets, so it might turn out to be dirty or impossible without making a PR to python itself, which is a whole other can of worms...

svinota:
@etene JFYI: I plan to keep Python compatibility back to version 3.9 as for now. Will be there any issues with that — pls let me know.

etene:
> @etene JFYI: I plan to keep Python compatibility back to version 3.9 as for now. Will be there any issues with that — pls let me know.

@svinota I'm pretty sure i can make it work with 3.9 upwards, my initial draft was based on 3.12 out of habit and comfort but the really important asyncio parts have been there since 3.7.

svinota:
The reason behind 3.9 (FIXME: I should document it) is `socket.recv_fds()` that is not supported in earlier versions — and which is used in the core functionality to init netns-enabled sockets.

svinota:
@etene regarding the tests I would propose this thing:

1. save binary packets (optionally, store them as hex dumps, see [1])
2. use `config.mock_dhcp` in a way like [2]
3. having this, just mock the traffic through the socket, and assert every state transition
4. bonus: now code snippets in the docs can be tested as well, to keep the docs up to date automatically, like in [3][4]

[1] https://github.com/svinota/pyroute2/blob/899fc581947f235343a9c6ef575ff21991ab786a/tests/test_core/test_ipr/test_route_dump_async.py#L7
[2] https://github.com/svinota/pyroute2/blob/899fc581947f235343a9c6ef575ff21991ab786a/pyroute2/netlink/rtnl/iprsocket.py#L93
[3] source: https://github.com/svinota/pyroute2/blob/19a309be953f6dd7100178188f764334deb1909f/pyroute2/iproute/linux.py#L254
[4] rendered: https://pyroute2.org/pyroute2-0.8.1-109-g19a309be/iproute_linux.html#pyroute2.iproute.linux.RTNL_API.filter_messages

tests/test_linux/fixtures/dhcp_servers/udhcpd.py
```diff
@@ -0,0 +1,101 @@
...
+        return "\n".join(f"{opt}\t{value}" for opt, value in lines)
+
+    async def __aexit__(self, *_):
+        await super().__aexit__(*_)
+        self._temp_dir.__exit__(*_)
+
+    def get_cmdline_options(self) -> tuple[str]:
+        '''All commandline options passed to udhcpd.'''
+        return (
+            'udhcpd',
```
svinota, line 71: use `busybox udhcpd` instead, on Fedora there is no separate link for that, unlike Ubuntu; but running the plugin this way as `busybox udhcpd` works on both Fedora and Ubuntu.

etene: It is actually what's used, when building the command line the fixture uses `BINARY_PATH` and the rest (from `get_cmdline_options`) is just its args.

etene:
@svinota I read your comment about using hex dumps for tests, that is a great idea for properly unit testing everything. Actually I was thinking about going as far as to store these test packets in pcap files, which might be easier for debugging.

However if you don't mind i would still like to have some real integration tests with a couple of real DHCP servers.

svinota:
@etene sure thing, leave the integration in place, it isn't mutually exclusive.

pyroute2/dhcp/messages.py
```diff
@@ -0,0 +1,151 @@
...
+    '''Make a REQUEST for an existing lease.
+
+    This differs from REQUESTs in response to an OFFER in that the server_id
+    option is never set.
+
+    When rebooting, the message is broadcast, and the requested_ip option is
+    set to the IP in the stored lease. The bootp client IP is left blank.
+
+    When renewing, (i.e. T1 expires) the message is for the server that granted
+    the lease. The leases's IP is expected to be assigned to the client's
```
BrianBaboch, line 101:
```suggestion
    the lease. The lease's IP is expected to be assigned to the client's
```

pyroute2/dhcp/messages.py
```diff
@@ -0,0 +1,151 @@
...
+    interface at this point.
+
+    When rebinding (T2), the message is broadcast on the network.
+
+    In both cases, the bootp client IP (ciaddr) is set to the leases's IP.
```
BrianBaboch, line 106:
```suggestion
    In both cases, the bootp client IP (ciaddr) is set to the lease's IP.
```

svinota:
Rebasing and merging into `p10-dhcp-protocol` branch.

Lets continue in one repo.

PR #1256, opened 2025-01-24
ifinfmsg: fix xdp flags

Farbey-1:
The kernel expects a 4-byte in host-endian format, but '>H' in Python sends a 2-byte in big-endian format.
https://github.com/torvalds/linux/blob/21266b8df5224c4f677acf9f353eecc9094731f0/net/core/rtnetlink.c#L3329


svinota:
Thanks a lot, merged!

PR #1257, opened 2025-01-24
Add hooks and other improvements

BrianBaboch:


pyroute2/dhcp/client.py
```diff
@@ -211,7 +211,7 @@ async def _expire_lease(self):
         self.timers._reset_timer('expiration')
         self.state = fsm.State.INIT
         # FIXME: call hooks in a non blocking way (maybe call_soon ?)
-        for i in self.config.hooks:
+        for i in reversed(self.config.hooks):
```
svinota, line 214: Just out of curiosity: why reversed?

pyroute2/dhcp/hooks.py
```diff
@@ -24,21 +25,54 @@ async def unbound(self, lease: Lease):
...
+            )
+
+    async def unbound(self, lease: Lease):
+        if lease.default_gateway is None:
+            LOG.error("Lease doesn't contain default gateway")
+            return
+        LOG.debug('Removing %s as default route.', lease.default_gateway)
+        async with AsyncIPRoute() as ipr:
+            ifindex = await ipr.link_lookup(ifname=lease.interface)
+            await ipr.route(
```
svinota, line 73: Can it possible that the route will be removed automatically before this hook gets called? By the kernel or by some 3rd party software like NetworkManager.

Because then `ip.route('del', …)` will raise a `NetlinkError(code=errno.ENOENT)` here.

pyroute2/dhcp/leases.py
```diff
@@ -93,22 +94,45 @@ def ip(self) -> str:
...
+    def routers(self) -> list[str]:
+        return self.ack['options'].get('router', [])
+
+    @property
+    def default_gateway(self) -> str | None:
+        '''The default gateway for this interface.
+
+        As mentioned by the RFC, the first router is the most prioritary.
+        '''
+        return self.routers[0] if self.routers is not None else None
```
svinota, line 116: Is it possible to get here an instance of a type other than `(tuple, list, None)`?

svinota:
@etene beside of `ip.route('del', …)` LTGM, but review pls too

pyroute2/dhcp/client.py
etene: Sometimes we have to run a hook before another when we get a new lease; for example setting the gateway address depends on having the IP assigned to the interface. But when releasing the lease, we have to remove the gateway before removing the IP or it will fail (because the gateway will already have been implicitly removed when deleting the IP from the interface). That does not actually cause any problems except for spurious errors, but these are better avoided.

pyroute2/dhcp/hooks.py
etene: Yes, a try/except with a warning log would be great here.

pyroute2/dhcp/dhcp4msg.py
```diff
@@ -40,6 +40,7 @@ class dhcp4msg(dhcpmsg):
         (Option.COOKIE_SERVER, 'ip4list'),
         (Option.LPR_SERVER, 'ip4list'),
         (Option.HOST_NAME, 'string'),
+        (Option.MTU_SIZE, 'be16'),
```
etene, line 43: I think simply `MTU` would be enough since it's already a size in itself. Or maybe `INTERFACE_MTU` since it's the name in RFC 2132.

pyroute2/dhcp/hooks.py
```diff
@@ -24,21 +25,54 @@ async def unbound(self, lease: Lease):
...
     async def bound(self, lease: Lease):
-        LOG.info('STUB: add %s to %s', lease.ip, lease.interface)
-        # await ip(
-        #     "addr",
-        #     "replace",
-        #     f"{lease.ip}/{lease.subnet_mask}",
-        #     "dev",
-        #     lease.interface,
-        # )
+        LOG.debug('Adding IP %s to %s', lease.ip, lease.interface)
```
etene, line 28: I'd remove `IP` (since it's obvious that it is an IP when reading the log), and add the subnet mask. Maybe the log level would be better as INFO since it's pretty much doing the most important thing we expect from a DHCP client.

pyroute2/dhcp/hooks.py
```diff
@@ -24,21 +25,54 @@ async def unbound(self, lease: Lease):
...
+        async with AsyncIPRoute() as ipr:
+            await ipr.addr(
+                "replace",
+                index=await ipr.link_lookup(ifname=lease.interface),
+                address=lease.ip,
+                mask=lease.subnet_mask,
```
etene, line 34: Although we should definitely add some code to prevent the whole client from crashing if a hook fails, it would also be better to be more defensive (at least raising an exception with some context) for predictable failures like a missing option. Like for example a useful error message if the subnet mask is not provided.

pyroute2/dhcp/hooks.py
```diff
@@ -24,21 +25,54 @@ async def unbound(self, lease: Lease):
...
+                broadcast=lease.broadcast_address
+            )
+
+
+class ConfigureDefaultRoute(Hook):
+    async def bound(self, lease: Lease):
+        if lease.default_gateway is None:
+            LOG.error("Lease doesn't contain default gateway")
```
etene, line 53: Since it's a DHCP option, mentioning it might be clearer, like "Lease does not set the blah option"

pyroute2/dhcp/hooks.py
```diff
@@ -24,21 +25,54 @@ async def unbound(self, lease: Lease):
...
+            return
+        LOG.debug('Adding %s as default route through %s',
```
etene, line 55: Maybe should be at the info level too.

pyroute2/dhcp/hooks.py
```diff
@@ -24,21 +25,54 @@ async def unbound(self, lease: Lease):
...
+            await ipr.addr(
+                "del",
+                index=await ipr.link_lookup(ifname=lease.interface),
+                address=lease.ip,
+                mask=lease.subnet_mask,
+                broadcast=lease.broadcast_address
+            )
+
+
+class ConfigureDefaultRoute(Hook):
```
etene, line 50: If we only ever run hooks for bound/unbound states, i'm starting to think we could just use an async context manager, like:

```python
@asynccontextmanager
async def configure_ip(lease: Lease):
    # ip addr replace
    yield
    # ip addr del
```

But we would have to check first if there are no other cases where we want to call a hook.

@BrianBaboch what do you think ?

pyroute2/dhcp/leases.py
etene: I think the condition is wrong anyway; in cases where the `routers` option wasn't provided, `self.routers` will return `[]`, and `self.routers[0]` will raise `IndexError` because the code checks for `None` explicitly.

pyroute2/dhcp/hooks.py
```diff
@@ -24,21 +29,63 @@ async def unbound(self, lease: Lease):
...
+            ifindex = await ipr.link_lookup(ifname=lease.interface)
+            try:
+                await ipr.route(
+                    "del",
+                    dst="0.0.0.0/0",
+                    gateway=lease.default_gateway,
+                    oif=ifindex
+                )
+            except NetlinkError as err:
+                if err.code=errno.ESRCH:
```
etene, line 89: Shouldn't that be an `==` ?

I also think we should raise for other cases

pyroute2/dhcp/leases.py
```diff
@@ -93,22 +94,45 @@ def ip(self) -> str:
         return self.ack['yiaddr']
 
     @property
-    def subnet_mask(self) -> str:
+    def subnet_mask(self) -> int | None:
```
etene, line 97: Use Optional, we need to be compatible with py39

pyroute2/dhcp/hooks.py
```diff
@@ -24,21 +29,63 @@ async def unbound(self, lease: Lease):
...
     async def bound(self, lease: Lease):
-        LOG.info('STUB: add %s to %s', lease.ip, lease.interface)
-        # await ip(
-        #     "addr",
-        #     "replace",
-        #     f"{lease.ip}/{lease.subnet_mask}",
-        #     "dev",
-        #     lease.interface,
-        # )
+        LOG.info('Adding %s/%s to %s', lease.ip,
```
svinota, line 32: Not for this PR, just an idea for some future refactoring — go through the code and make all the logging in one single parsable format, be it golang `key=value` or json or anything else, but that could have single handler in the log collection stacks like loki/grafana or victorialogs. (looking is there any "make a ticket from this comment" :)

pyroute2/dhcp/hooks.py
```diff
@@ -24,21 +29,63 @@ async def unbound(self, lease: Lease):
...
+                gateway=lease.default_gateway,
+                oif=ifindex
+            )
+
+    async def unbound(self, lease: Lease):
+        if lease.default_gateway is None:
+            LOG.error('Lease does not set the router option')
+            return
+        LOG.info('Removing %s as default route.', lease.default_gateway)
+        async with AsyncIPRoute() as ipr:
```
svinota, line 79: Again, not a request to change the code.

@etene would we benefit if I add an option like "idempotent" (default False) to `AsyncIPRoute` — that would ignore `EEXIST` for `add` and `ENOENT` for `del` calls?

pyroute2/dhcp/hooks.py
etene: We do that in production, but it does not necessarily entail going through all logging calls. We use the "logfmt" format with a [custom logging formatter](https://github.com/RyaxTech/logfmt-logger) and that does the trick, we can export them to loki.

Since pyroute2 is a library, i believe logging format should be up to the callers anyway.

pyroute2/dhcp/hooks.py
etene: I think the most useful thing we could do in that regard would be to raise specific exceptions for errno codes, and make them inherit `NetlinkError`. That way, we wouldn't break any existing code while still allowing callers to catch specific exceptions instead of comparing the error code manually.

svinota: A good idea, thanks.

svinota:
Are we ready to merge it?

PR #1262, opened 2025-01-29
[WIP] dhcp: use composition in AsyncDHCP4Socket

svinota:
Don't inherit from AsyncRawSocket class, but instead use it as a property.

The idea is to ease AsyncRawSocket replacement with a mock class (the commit with such mock class will follow) to run unit testing.


svinota:
@etene , I will rebase & rework it once we merge the main dhcp code.

PR #1264, opened 2025-02-03
unify test fixtures

svinota:


svinota:
@etene here's a proposal to simplify the fixtures.

As soon as we agree on some fixtures layout, I will try to unify existing modules to use same fixtures across all the CI.

tests/test_core/conftest.py
```diff
@@ -75,15 +16,66 @@ async def p9(request, tmpdir):
...
+        ipr.poll(ipr.addr, 'dump', address='127.0.0.1', timeout=5)
+    yield ns
+    try:
+        netns.delete(ns)
+    except Exception:
+        pass
+
+
+@pytest.fixture
+def link(nsname):
```
etene, line 34: Type annotations for test fixtures, even if they're not checked (yet !), really help me understand what they do at a glance even when there's no doc.

etene, line 34: I  think fixture names should be a bit more explicit, so less guessing when trying to figure later what tests do

tests/test_core/conftest.py
```diff
@@ -75,15 +16,66 @@ async def p9(request, tmpdir):
...
+                raise
+
+
+@pytest.fixture
+def index(link):
+    yield link.get('index')
+
+
+@pytest.fixture
+def ifname(link):
```
etene, line 53: This fixture's name is too close to the `uifname` function, which is also used in tests, which might be confusing

tests/test_core/conftest.py
```diff
@@ -75,15 +16,66 @@ async def p9(request, tmpdir):
...
+        yield link
+        try:
+            ipr.link('del', index=link.get('index'))
+        except NetlinkError as e:
+            if e.code != errno.ENODEV:
+                raise
+
+
+@pytest.fixture
+def index(link):
```
etene, line 48: The name here is also too generic imo

tests/test_core/conftest.py
```diff
@@ -75,15 +16,66 @@ async def p9(request, tmpdir):
...
+    yield link.get('index')
+
+
+@pytest.fixture
+def ifname(link):
+    yield link.get('ifname')
+
+
+@pytest.fixture
+def tmp_link(nsname):
```
etene, line 58: i guess there should be a sync/async version of each relevant fixture ? just a thought, i have no real idea about naming/separation yet

tests/test_core/conftest.py
svinota: oh, you urge me to use annotations

ok, I'll try to change my habits :)

tests/test_core/conftest.py
svinota: Nope, there is no need to have both sync/async versions of fixtures.

The only reason why I use sync/async pair for IPRoute is because I need to test both APIs; otherwise one version is enough, and one can mix them

etene: since the dhcp tests i wrote are async, i had a preference for async fixtures (where the tests are async). Especially because mixing blocking sync code & async code can lead to pretty bad surprises, like:

```python
import asyncio
import time

async def its_a_trap():
    before = time.time()
    async with asyncio.timeout(1):
        time.sleep(2)
    print('took', int(time.time() - before), 'secs !')
```

(as you probably already know it takes 2 seconds.  and I have a lot of asyncio timing stuff in the client :D)

svinota: A good point.

So far I needed only guaranteed states in the test fixtures, like "test interface created and accessible", "test netns created and set for the IPRoute" etc. Thus blocking synchronous fixtures, and the only asynchronous fixture is the one that provides AsyncIPRoute.

But that's something I have to review.

tests/test_core/conftest.py
svinota: Would `test_if_index` work?

tests/test_core/conftest.py
etene: Ok ! Understood. I think it's not needed to implement a whole set of async fixtures for now, just properly name/write fixtures so that adding async equivalents later won't be a pain

tests/test_core/conftest.py
etene: i don't think it's necessary to put `test` in pytest fixture names (pytest itself does not do it). However the interface for which this fixture returns an ifindex is always a dummy, isn't it ? So maybe the fixtures could be named `dummy_ifindex`, `dummy_ifname`, `dummy_interface`... ?

tests/test_core/conftest.py
```diff
@@ -75,15 +16,66 @@ async def p9(request, tmpdir):
...
+            if e.code != errno.ENODEV:
+                raise
+
+
 @pytest_asyncio.fixture
-async def async_ipr(request, tmpdir):
+async def async_ipr(nsname, request):
     kwarg = getattr(request, 'param', {})
-    async with AsyncIPRouteContext(**kwarg) as ctx:
+    async with AsyncIPRoute(netns=nsname, **kwarg) as ctx:
```
etene, line 73: The `async_ipr` fixture is exactly what i need.

However i quite liked the idea of having a custom fixture class like `AsyncIPRouteContext`. That would allow us to add async test utility methods on it. I would find it really useful to be able to:

- `assert await async_ipr.address_exists(addr)`,
- `assert await async_ipr.interface_exists(iface)`,
- `assert await async_ipr.route_exists(rt)`,

And so forth. 
As a bonus i guess it would make it unnecessary to pass the netns to these because the fixture would already have the info.
It would also reduce the number of imports & "free-floating" test functions.

What do you think ?

svinota: 1. One thing I definitely wanna stick to, is to separate the test object (any pyroute2 object) and the assertion routines.

I had a really bad experience with `tests/test_linux` and its fixtures, when rewriting the netlink protocol internals I effectively stopped all the CI, making it impossible to locate the issue — and I had to switch back to the manual testing for a month or like that.

Thus I created these functions: https://github.com/svinota/pyroute2/blob/master/tests/net_tools.py . The functions are not as convenient to use, and there is a lot of things to improve, but they can be used for some basic assertions, and don't use any pyroute2 functionality. 

I think we can provide them (the assertion functions) in the namespace of the fixture context, like you said, but I would never again use the test subject in the assertion tooling.

2. I believe that the assertions must be synchronous — "now stop everything, make a probe, analyze, continue"

svinota: But I also see a pro in having assertions as free-floating — in the imports section of the test module I can see, what the assertions are used there.

svinota:
@etene so here is the final version of the proposal.

1. make fixtures a library module in the main tree
2. provide them as pytest plugins

So it will be possible to use the fixtures outside of the project through the same plugin declaration in the `conftest.py`.

The docs will be ready by Sunday, but the API is ready now.

The fixtures provide two ways:

1. either use `sync_context` and `async_context`, and have `ipr`, `netns` and `test_link` as properties of the context
2. or use the fixtures separately, which makes possible to control fixtures consumption in the test signatures.

tests/test_core/conftest.py
etene: That seems completely reasonable, thanks for the context !

svinota:
@etene take a look at the code & docs

The only thing that makes me worry is that having `import pytest` in the library source may trigger some weird side-effects in the build environments for rpm or deb that use dependencies autodetection. I'll try to dig deeper a bit before merging.

pyroute2/fixtures/iproute.py
```diff
@@ -0,0 +1,523 @@
...
+
+from pyroute2 import NDB, netns
+from pyroute2.common import uifname
+from pyroute2.iproute.linux import AsyncIPRoute, IPRoute
+from pyroute2.netlink.exceptions import NetlinkError
+from pyroute2.netlink.rtnl import IFNAMSIZ
+from pyroute2.netlink.rtnl.ifinfmsg import ifinfmsg
+
+
+class TestInterface:
```
etene, line 16: Awesome & annotated, this kind of fixture is perfect !

pyroute2/fixtures/iproute.py
```diff
@@ -0,0 +1,523 @@
...
+    '''The test context.
+
+    Provided by `async_context` and `sync_context` fixtures.
+
+    Provides convenient shortcuts to RTNL API, the network namespace
+    name and the test interface spec.
+    '''
+
+    def __init__(
+        self, ipr: Union[IPRoute, AsyncIPRoute], test_link: TestInterface
```
etene, line 78: This fixture class would benefit from being `Generic[IPRouteT]` to avoid confusing the type checker later but i can try to do it later if you want.

pyroute2/config/__init__.py
```diff
@@ -4,6 +4,36 @@
 import signal
 import socket
 
+
+class LocalMock:
```
etene, line 8: Wouldn't an stdlib Mock provide the same functionality ?

pyroute2/fixtures/iproute.py
```diff
@@ -0,0 +1,523 @@
...
+                name='nsname'
+            ):
+                nsptr = nsname
+                calls += 1
+                test_list_interfaces(nsname)
+
+        call_create = netns.create().call_args_list
+        call_remove = netns.remove().call_args_list
+        nslist_create = set([x[0][0] for x in call_create if x[0]])
+        nslist_remove = set([x[0][0] for x in call_remove if x[0]])
```
etene, line 154: ok i might be nitpicking here but this looks a bit complicated for example code (maybe comments or a simpler case ?)

pyroute2/fixtures/iproute.py
```diff
@@ -0,0 +1,523 @@
...
+        assert calls == 1
+        assert nsptr is not None
+        assert nsptr in nslist_create
+        assert nsptr in nslist_remove
+    '''
+    nsname = uifname()
+    netns.create(nsname)
+    with IPRoute(netns=nsname) as ipr:
+        ipr.link('set', index=1, state='up')
+        ipr.poll(ipr.addr, 'dump', address='127.0.0.1', timeout=5)
```
etene, line 165: Oh, if this does what i think it does, it's really interesting & i did not know about it ! (maybe add a comment ?)

pyroute2/fixtures/iproute.py
```diff
@@ -0,0 +1,523 @@
...
+            await async_ipr.link('set', index=1, state='up')
+
+    .. testcode::
+        :hide:
+
+        patcher.stop()
+
+        import asyncio
+
+        async def test_wrapper(calls):
```
etene, line 369: this function clearly needs a bit of explaining imho :)

pyroute2/config/__init__.py
svinota: Yes. 

But there is an unspoken requirement :) The library core must use as little of the stdlib as possible, since it runs sometimes on embedded platforms with limited stdlib. That's why we have `pyroute2-minimal` with `IPRoute` + `netns` functionality only.

An unittest is one of the modules that gets cut when they build embedded platforms.

pyroute2/fixtures/iproute.py
svinota: Notice `:hide:`

It's not an example code, it's an assertion code, it will not be shown in the rendered docs.

But you have right, it's better to move this assertion outta there not to confuse plain old `help()` users.

pyroute2/fixtures/iproute.py
svinota: It waits for a match in the dump output.

https://docs.pyroute2.org/iproute_linux.html#pyroute2.iproute.linux.RTNL_API.poll

pyroute2/fixtures/iproute.py
svinota: That's also an assertion code, not an example.

Yep, you have convinced me, I'll rearrange the assertion code so it will not be visible in the docstrings.

pyroute2/fixtures/iproute.py
svinota: Done.

PR #1265, opened 2025-02-04
nlsocket: lower log level when a port is already used

JeThWifirst:
When a port is already used or the socket cannot be reused, it is expected to have an Exception, no need to log an error for this case.

And in case all ports are exhausted, an Exception will be raised.


svinota:
Agree. Thanks, merging!

PR #1267, opened 2025-02-06
dhcp client: add missing default values for rebinding & renewal times

etene:
I had overlooked a critical part of the RFC and forgot to compute default values for renewal & rebinding times; not all servers provide these. So the lease expired again and again.

### DIFF
--- a/pyroute2/dhcp/leases.py
+++ b/pyroute2/dhcp/leases.py
@@ -2,11 +2,11 @@
 
 import abc
 import json
-import random
 import time
 from dataclasses import asdict, dataclass, field
 from logging import getLogger
 from pathlib import Path
+from secrets import SystemRandom
 from typing import Generic, Optional, TypeVar
 
 from pyroute2.common import dqn2int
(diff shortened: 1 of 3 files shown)


pyroute2/dhcp/leases.py
```diff
@@ -97,26 +98,32 @@ def expiration_in(self) -> Optional[float]:
...
-        Can be negative if it's past due,
-        or `None` if the server didn't give a rebinding time.
+        Can be negative if it's past due.
         '''
-        return self._seconds_til_timer('rebinding')
+        if lease_rebinding := self._seconds_til_timer('rebinding'):
+            return lease_rebinding
+        # RFC section 4.4.5 says we need a fuzzy value around 0.875
+        # FIXME will crash if there is no lease time
+        return self.expiration_in * random.uniform(0.75, 0.90)
```
svinota, line 126: Don't want to seem picky, but probably it would be nice to use SystemRandom consistently?

etene: of course, my bad

svinota:
Thanks!

PR #1271, opened 2025-02-07
dhcp: improve & properly test hooks

etene:
There are lots of ways for dhcp hooks to fail, they need more tests


tests/test_linux/conftest.py
```diff
@@ -20,6 +16,7 @@
 
 config.nlm_generator = True
 pytest_plugins = 'pytester'
+pytest_plugins = ["pyroute2.fixtures.iproute"]
```
svinota, line 19: pls don't leave `pytester` out of the list

etene: oops !

svinota:
Thanks, merged!

PR #1272, opened 2025-02-10
Delete unused code from `./common.py`

skatromb:
Deletion of unused code in `pyroute2/common.py`:

- Python 2 related chunk
- Classes: `View` and `DotKeys`
- Functions: `hexload`, `_fnv1_python2` and `failed_class`
- Constants: `AF_PIPE` and `DEFAULT_RCVBUF`

Take extra attention to deletion of:

- `AddrPool.allocated`
- `AddrPool.alloc_multi`
- `AddrPool.setaddr`

...because I hope I checked usage of these attributes, however, I am not 100% sure

### DIFF
--- a/pyroute2/common.py
+++ b/pyroute2/common.py
@@ -285,7 +285,6 @@ class AddrPool(object):
         mx = self.cell
         self.reverse = reverse
         self.release = release
-        self.allocated = 0
         if self.release and not isinstance(self.release, int):
             raise TypeError()
         self.ban = []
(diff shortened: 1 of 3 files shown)


pyroute2/common.py
```diff
@@ -582,10 +390,7 @@ def _fnv1_python3(data):
     return hval & 0xFFFFFFFF
 
 
-if sys.version[0] == '3':
-    fnv1 = _fnv1_python3
-else:
-    fnv1 = _fnv1_python2
+fnv1 = _fnv1_python3
```
svinota, line 393: simply rename the function:

```
-- def _fnv1_python3(…
++ def fnv(…
```

skatromb: Thanks, done

Issue #1266, opened 2025-02-04
Local opened event loop are not closed


ffourcot:
Hello,

We have a lot of resource warning for unclosed event loop.

I had a short look on the code, and it looks like loop opened here in case of RuntimeError:

```python
      def setup_event_loop(self, event_loop=None):
          if event_loop is None:
              try:
                  event_loop = asyncio.get_running_loop()
                  self.status['event_loop'] = 'auto'
              except RuntimeError:
                  event_loop = asyncio.new_event_loop()
                  self.status['event_loop'] = 'new'
          return event_loop
```

is never closed here:

```python
          for (
              sock,
              msg_queue,
              event_loop,
              transport,
              protocol,
          ) in self.__all_open_resources:
              event_loop.call_soon_threadsafe(send_terminator, msg_queue)
              transport.close()
              if sock is not None:
                  sock.close()
          self.__all_open_resources = tuple()
```

What do you think to close event loop at the end, if it is in "new" status?

ffourcot:
Other question: you chose to use get_running_loop instead of get_event_loop. It means that you will create a loop for any non-async code, even if not necessary (the loop can exists, but not currently running). What is mandatory to provide compatibility code?

svinota:
Thanks for the heads-up, I have to review this code.

Regarding the other question, it was done because `get_running_loop` has much more simple and predictable behaviour.

The plan was that if someone wants to use an existing event loop, they could provide it as an argument to the socket constructor; but it looks that I missed this part of the functionality, so again, thanks a lot for the heads-up.

Marking this ticket as a bug, to be fixed asap. If you have proposals regarding the event loop management — pls comment, I will try to align the code with your use cases.

inemajo:
Hello,

What do you think of using a custom event loop policy, which would return a event loop by default ? something like that:

```diff
diff --git a/pyroute2/netlink/core.py b/pyroute2/netlink/core.py
index f437a180..1989c0f4 100644
--- a/pyroute2/netlink/core.py
+++ b/pyroute2/netlink/core.py
@@ -25,6 +25,20 @@ CoreSocketResources = collections.namedtuple(
 )


+class DefaultEventLoop(asyncio.DefaultEventLoopPolicy):
+
+    _default_event_loop = None
+
+    def get_event_loop(self):
+        if self._default_event_loop is None:
+            self._default_event_loop = self.new_event_loop()
+            asyncio.set_event_loop(self._default_event_loop)
+        return super().get_event_loop()
[...]

PR #1277, opened 2025-02-11
core: use_event_loop option

svinota:
Event loop initialization and retrieval.

Bug-Url: https://github.com/svinota/pyroute2/issues/1266

### DIFF
--- a/pyroute2/netlink/core.py
+++ b/pyroute2/netlink/core.py
@@ -182,6 +182,7 @@ class AsyncCoreSocket:
         self.status = self.spec.status
         self.request_proxy = None
         self.local = threading.local()
+        self.lock = threading.Lock()
         if use_event_loop:
             self.local.event_loop = use_event_loop
             self.status['use_event_loop'] = True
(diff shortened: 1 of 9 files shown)


pyroute2/netlink/core.py
```diff
@@ -165,12 +165,14 @@ def __init__(
         flags=os.O_CREAT,
         libc=None,
         groups=0,
+        use_event_loop=False,
```
ffourcot, line 168: perhaps that this option should be accepted in IPRoute? We have a lot of code using the "sync" class

ffourcot: I was two minutes too late

pyroute2/netlink/core.py
```diff
@@ -234,9 +241,22 @@ def connection_lost(self):
 
     @property
     def event_loop(self):
+        return self.ensure_event_loop()
+
+    def ensure_event_loop(self):
         if not hasattr(self.local, 'event_loop'):
+            if self.status['use_event_loop']:
+                if self.status['use_thread_id'] == id(
```
ffourcot, line 249: s/use_thread_id/thread_id ?

pyroute2/netlink/core.py
svinota: Good catch, thanks! And it's never too late :)

I will add that definitely.

PR #1280, opened 2025-02-13
iproute: use_event_loop in sync version

svinota:
Bug-Url: https://github.com/svinota/pyroute2/issues/1266
Bug-Url: https://github.com/svinota/pyroute2/pull/1277

### DIFF
--- a/pyroute2/netlink/core.py
+++ b/pyroute2/netlink/core.py
@@ -184,7 +184,7 @@ class AsyncCoreSocket:
         self.local = threading.local()
         self.lock = threading.Lock()
         if use_event_loop:
-            self.local.event_loop = use_event_loop
+            self.status['event_loop'] = id(use_event_loop)
             self.status['use_event_loop'] = True
             self.status['thread_id'] = id(threading.current_thread())
         if libc is not None:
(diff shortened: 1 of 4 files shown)


pyroute2/netlink/core.py
```diff
@@ -184,9 +184,9 @@ def __init__(
         self.local = threading.local()
         self.lock = threading.Lock()
         if use_event_loop:
-            self.local.event_loop = use_event_loop
             self.status['use_event_loop'] = True
             self.status['thread_id'] = id(threading.current_thread())
+            self.status['event_loop'] = id(use_event_loop)
```
ffourcot, line 189: I do not understand this line, why are you calling id() on a boolean?

svinota: Oh lol.

svinota: Updated. Looks like status struct needs some typization. Give me some time to go through it tonight (but this version is already working)

svinota: Shortly speaking, `status['use_event_loop']` might be either `False` or `True`. If it is `True` then we're using a predefined event loop, and must reject to work outside of the initial thread.

If it is `False`, then we try to get a running event loop from the current thread (`status['event_loop'] == 'auto'`), or create a new one (`status['event_loop'] == 'new'`).

Probably there might be a better way to express this choice.

svinota: Yep, let me write it over with typization.

svinota:
Okay, the config dataclass will come as a separate PR, there is too much work TBH, and I would not like to keep this particular feature hanging for a long time in the WIP state.

@ffourcot if no objections, I would merge it today.

ffourcot:
It looks good for me, thanks

PR #1281, opened 2025-02-13
netlink/core: give a chance for socket to close

ffourcot:
transport.close() does not really close the socket, but move the transport to "closing" state.

We need to give some time to the event loop for a real closed object

Before the patch:

$ python3 -X dev -c "from pyroute2 import IPRoute ; ip = IPRoute() ; ip.close()"
/usr/lib/python3.11/asyncio/selector_events.py:843: ResourceWarning: unclosed transport <_SelectorDatagramTransport closing fd=6>


svinota:
Something weird on the CI, going to debug later today

PR #1282, opened 2025-02-13
dhcp: refactor typing, ip list decoding bug

etene:
mypy does not return errors anymore for the dhcp module, and i found a bug in the process

### DIFF
--- a/pyroute2/dhcp/__init__.py
+++ b/pyroute2/dhcp/__init__.py
@@ -101,6 +101,15 @@ the code 255 is the end of options code.
 import struct
 import sys
 from array import array
+from typing import (
+    Any,
+    Callable,
+    NamedTuple,
+    Optional,
+    TypedDict,
+    TypeVar,
+    Union,
+)
 
 from pyroute2.common import basestring
 from pyroute2.protocols import _decode_mac, _encode_mac, msg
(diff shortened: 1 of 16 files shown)


pyroute2/dhcp/__init__.py
```diff
@@ -110,25 +119,33 @@ class array8(option):
 if not hasattr(array, 'tobytes'):
     # Python2 and Python3 versions of array differ,
     # but we need here a consistent API w/o warnings
-    class array(array):
-        tobytes = array.tostring
+    # FIXME: we don't support py2, probably not needed
```
etene, line 122: @svinota still needed ?

pyroute2/protocols/__init__.py
```diff
@@ -118,9 +118,12 @@ def _encode_mac(value: str) -> list[int]:
 
 
 class msg(dict):
-    buf = None
+    # FIXME: figure out why these are declared as class attributes
```
etene, line 121: @svinota could you tell me why there seem to be instance variable declared as class attributes in some places ?

pyroute2/dhcp/__init__.py
svinota: Nope. Should be safe to remove.

pyroute2/protocols/__init__.py
svinota: No particular reason. Just remnants of prototyping.

PR #1285, opened 2025-02-14
dhcp client: add signal handlers to renew/rebind/reset the lease

etene:
- SIGUSR1 triggers a renewal
- SIGUSR2 triggers a rebinding
- SIGHUP resets the client


svinota:
Looks ok, merging. Thanks a lot!

PR #1286, opened 2025-02-14
Types for the `pyroute2.common`

skatromb:
Couldn't infer type for `msg_done` function.

Linter would warn about two lambda functions, but I saved them, since one of them is actually closure, so that would complicate typing too much.

Other than typing, following changes were done:
- `map_namespace` were re-written to be more explicit. Tests are added
- `load_dump` Python 2 code was removed

### DIFF
--- a/pyroute2/common.py
+++ b/pyroute2/common.py
@@ -7,7 +7,6 @@ import io
 import os
 import socket
 import struct
-import sys
 import threading
 import time
 import types
--- a/pyroute2/common.py
+++ b/pyroute2/common.py
@@ -205,7 +205,7 @@ def dqn2int(mask: str, family: socket.AddressFamily = socket.AF_INET) -> int:
     return ret
 
 
-def get_address_family(address):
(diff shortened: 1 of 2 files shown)


svinota:
@skatromb I convert it into a draft until we get it working :)

pyroute2/common.py
```diff
@@ -71,11 +72,13 @@
 # General purpose
 #
 class Namespace(object):
-    def __init__(self, parent, override=None):
+    def __init__(
+        self, parent: 'Namespace', override: Union[dict[str, Any], None] = None
```
etene, line 76: I think `Optional[...]` is slightly better and less verbose than `Union[..., None]` for these cases. And it will ease refactoring to  `... | None` when dropping support for 3.9

skatromb: Great point, thank you! I forgot that it exists

etene: you're welcome !

PR #1290, opened 2025-02-17
dhcp: interface name can have a vlan in it

BrianBaboch:
The way it was done, if the interface name had a vlan in it (example eth1.100), since we're using the `with_suffix` function it would remove the vlan and save the files as `eth1.pid` and `eth1.lease.json`


svinota:
Oh, thanks

PR #1296, opened 2025-02-20
dhcp server detector: use a different xid per interface

etene:
We noticed an issue when using the dhcp server detector over a vlan, i have to test & reproduce it but using a different xid for each interface will help

### DIFF
--- a/pyroute2/dhcp/dhcp4socket.py
+++ b/pyroute2/dhcp/dhcp4socket.py
@@ -12,6 +12,7 @@ from typing import Optional
 from pyroute2.compat import ETHERTYPE_IP
 from pyroute2.dhcp.dhcp4msg import dhcp4msg
 from pyroute2.dhcp.messages import ReceivedDHCPMessage, SentDHCPMessage
+from pyroute2.ext.bpf import BPF
 from pyroute2.ext.rawsocket import AsyncRawSocket
 from pyroute2.protocols import ethmsg, ip4msg, udp4_pseudo_header, udpmsg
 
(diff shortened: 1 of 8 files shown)


tests/test_linux/test_dhcp/test_server_detector.py
```diff
@@ -143,3 +145,124 @@ async def shutdown_iface(ifindex: int):
...
+        # we stopped udhcpd so only got an offer from dnsmasq
+        assert len(offers) == 1
+
+    #################################################################
+    # Here is the issue:
+    # the socket listening on the non-vlan interface also receives
+    # the offer sent over the vlan. since we have a different xid per
+    # interface, it is discarded, but should not happen anyway.
+    assert b'Got OFFER with xid mismatch, ignoring' not in stderr
+    #################################################################
```
etene, line 268: This assert fails because the raw socket receives & parses without complaining VLAN tagged packets that are not supposed to be "meant for it". Filtering at the right place (BPF ?) should solve the issue.

etene: @svinota i think i found the source of the problem:

- since we have a raw packet socket, we get packets for vlans "on top" of the socket's interface too
- the linux kernel apparently drops the vlan info from the packet early on (even in [libpcap](https://github.com/the-tcpdump-group/libpcap/blob/3a7962215f4f3b13ac792ad0e0487a53c0f1178e/pcap-linux.c#L1791) they have to add it back)

So the solution is either
- to use the `PACKET_AUXDATA` socket option to get the vlan from the kernel in ancillary data,
- or to add a few instructions to the BPF filter. Fortunately [it can be done](https://lore.kernel.org/netdev/51FB6A9D.2050002@redhat.com/T/).

I'm going to use the BPF approach so the filtering is not done in userspace; i'll just drop all packets that have a vlan, so we only get packets for the expected interface everytime. I have a commit coming up as soon as it's cleaned up a little.

pyroute2/dhcp/dhcp4socket.py
```diff
@@ -20,22 +21,40 @@
...
 
 
 def listen_udp_port(port: int = 68) -> list[list[int]]:
-    # pre-scripted BPF code that matches UDP port
+    '''BPF filter that matches Ethernet + IPv4 + UDP on the given port.
+
+    Packets tagged on a vlan are also dropped, see
+    https://lore.kernel.org/netdev/51FB6A9D.2050002@redhat.com/T/
+    '''
     bpf_code = [
```
svinota, line 34: [not a request for changes] I wonder, if we add some basic BPF compiler one day :)))

tests/test_linux/test_dhcp/test_server_detector.py
```diff
@@ -143,3 +145,125 @@ async def shutdown_iface(ifindex: int):
     assert (
         f'{veth_pair.client!r}: [Errno 100] Network is down'
     ) in stderr.decode()
+
+
+@pytest.mark.parametrize(
+    ('run_dhcp_server_outside_vlan',), ((True,), (False,))
+)
+async def test_detect_with_vlan(
```
svinota, line 153: Excellent, thanks

pyroute2/dhcp/dhcp4socket.py
etene: While working on the filter i was thinking exactly the same thing, i hope to come across a use case that will require it one day :) for now having constants instead of raw int values is a first step, and you already wrote `compile()` in bpf.py which is arguably a very basic BPF compiler !

PR #1298, opened 2025-02-20
Async uevent

michaelweiser:
Thanks for this awesome library!

Here are some (hopefully self-explanatory) extensions to uevent handling. I use them to listen for backlight changes. Would you be interested in including those? What would potentially need doing to get them in?


svinota:
Great, thanks for the contribution!

Would be nice to update the PR:

* the code should pass `make format` w/o errors — actually, `make format` tries to reformat the code, see `git diff` after running `make format`
* [optional but something to think about] add some functional test, maybe using mock; see `/noxfile.py` and `/tests`
* [optional but really nice] add types annotation and include the affected file into `/.mypy-check-paths`

svinota:
There is an issue with the CI, so some jobs are failing; I'm to fix that tonight

michaelweiser:
`make format` is done but `mypy` causes some scope creep as it seems I need to put type annotations on all the superclasses as well to avoid "incompatible signature" errors. I'll play with it some more to see where it ends. :)

svinota:
The CI is fixed.

We can merge the code, and the annotations, and occasionally some tests you can push later as a separate PR.

michaelweiser:
Here's an attempt at typing. Way more than what I bargained for. :)

svinota:
Great! Thanks a lot!

PR #1299, opened 2025-02-22
Delete code that manages old Python versions, that are not supported anymore

skatromb:


pyroute2/loader.py
```diff
@@ -26,18 +26,4 @@
...
-            log.warning('patching struct.unpack_from()')
-
-            def wrapped(fmt, buf, offset=0):
-                return struct._u_f_orig(fmt, str(buf), offset)
-
-            struct._u_f_orig = struct.unpack_from
-            struct.unpack_from = wrapped
-        else:
-            raise
+    struct.unpack_from('I', bytearray((1, 0, 0, 0)), 0)
```
svinota, line 29: This call was just a probe to raise an exception in older python versions, and to prepare a patch for them.

Since we don't support these versions now, there is no need for this probe at all, this `init()` can be removed completely, and the `loader` module can be dropped.

Let me think a bit, how to that in the best possible way.

skatromb:
Please take a look at 
`pyroute2/loader.py:15:1: F401 'pyroute2.config.log' imported but unused`
, if that could be removed

svinota:
@skatromb I have removed the loader module as not used anymore.

PR #1300, opened 2025-02-25
dhcp: fix typing

etene:


pyroute2/dhcp/client.py
```diff
@@ -649,17 +641,15 @@ def _add_signal_handler(
...
         '''Arranges for `callback` to be called when receiving `signal`.
 
         `callback` must take a `signal` keyword argument.
         '''
         loop.add_signal_handler(
             sig=signal,
-            callback=lambda: asyncio.create_task(
-                callback(signal=signal)  # type: ignore[arg-type]
-            ),
+            callback=lambda: asyncio.create_task(callback(signal=signal)),
```
svinota, line 652: Oh, good.

Thanks.

pyroute2/dhcp/timers.py
```diff
@@ -49,7 +47,7 @@ def arm(self, lease: Lease, **callbacks: AsyncCallback) -> None:
         loop = asyncio.get_running_loop()
 
         for timer_name, async_callback in callbacks.items():
-            self._reset_timer(timer_name)  # type: ignore[arg-type]
+            self._reset_timer(timer_name)
```
svinota, line 50: Not sure if it is a good idea to add some CI test that raises warnings if there are any inline linter overrides.

etene: as long as it does not make the CI fail, i see no issue with it !

PR #1308, opened 2025-03-03
dhcp: add most if not all options to the enum

etene:


etene:
@svinota i added as many missing options as I could find. However i did not add the matching decoding code yet, because I noticed a format discrepancy for an option, which would make it fail to decode in certain cases. And I guess there would be other cases. I have to investigate that before writing the encode/decode stuff. I still have a test to write and then this will be mergeable for now !

svinota:
Thanks, then I wait for the test!

etene:
@svinota ready for merging !

PR #1316, opened 2025-03-07
dhcp: decode unknown options as list of ints instead of bytes

etene:
I made the ill-advised choice to decode unknown DHCP options as raw bytes. This works pretty well as long as you stay within Python but will causes issues when serializing or deserializing. Lists of ints are easy to use and don't cause such issues.


svinota:
Oke, thanks. Looks ok, merging

PR #1325, opened 2025-03-20
dhcp client: don't crash if the interface does not exist, print an error message instead

etene:


svinota:
@etene what's the reason is behind this decision?

etene:
> @etene what's the reason is behind this decision?

Since I didn't check properly for the interface to exist, if the client was started with an invalid interface, it would just crash and print the python traceback for an IndexError.

We needed to have a simple exception to catch in these cases for our uses, plus the cli now prints a useful error instead of a traceback with an obscure generic exception.

svinota:
Thanks a lot!

Issue #1336, opened 2025-04-25
epid attr compatibility issue


svinota:
> Currently, if we upgrade to pyroute2 >=0.9.1, we see errors with using
the package.  However, with pyroute2 >= 0.6 (tested with 0.7.3 and
0.8.1), I don't get see issues.  Ex:
```
  File "/home/core/git/linux/tools/testing/selftests/net/openvswitch/ovs-dpctl.py", line 1972, in create       
    procarray += [int(p.epid)]
  File "/usr/local/lib/python3.9/site-packages/pyroute2/netlink/core.py", line 712, in __getattr__
    raise AttributeError(key)
  AttributeError: epid
```

PR #1337, opened 2025-04-25
core: expose epid as an attribute

svinota:
In order to provide compatibility.

Bug-Url: https://github.com/svinota/pyroute2/issues/1336

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -37,7 +37,10 @@ jobs:
       - uses: actions/checkout@v4
       - run: make nox session=docs
   neutron:
-    runs-on: Linux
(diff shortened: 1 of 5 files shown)


apconole:
Note, in our testing we do have an upcall test (that's what uses the epid details), and with this patch, I see the following:

> TEST: upcall_interfaces                                             [START]
> TEST: ovs: test the upcall interfaces                               [ OK ]
> 

I need to check because I'm seeing a failure with the drop reasons test, but that could be related to my local environment.

apconole:
Drop reasons is due to my setup, it seems.  With this patch everything looks to be working.

svinota:
Great, thanks.

PR #1341, opened 2025-04-30
wireguard: fix peer flags

new-sashok724:
I didn't find where bit offsets for the WireGuard peer flags are converted into bitmasks, so I believe the values for these constants should have been bitmasks in the first place. I've prefixed all flag values with `0x` to make it clearer that they are bitmasks.

I discovered this issue when I noticed that the `wg.set` function overrides allowed IPs for the peer even when they were not provided, as if the `WGPEER_F_REPLACE_ALLOWEDIPS` flag (`0x02`) was set. It turned out that this was due to the `WGPEER_F_UPDATE_ONLY` flag (which was mistakenly defined as `2` before this change).


svinota:
Looks like the wireguard tests are failing. I will look tonight — most probably we have wrong tests as well.

new-sashok724:
Oh, I didn't notice that there were tests for WireGuard, sorry about that :sweat_smile:
I've taken a look, and it seems like the `WGPEER_F_UPDATE_ONLY` flag should not have actually been set, as peers are created by the tests. However, to preserve the ability to set that flag, I've added new peer options. This should fix the tests. Please take a look :slightly_smiling_face:

svinota:
Thanks a lot! Everything works as a charm.

Merging.

Issue #1343, opened 2025-05-06
Regression on generators and huge data


ffourcot:
Hello,

We are using the Conntrack class to read some values. A conntrack table can be huge, of course. It is why nlm_generator was set to True by default on this class.

However, it looks like nlm_generator has not effect at all anymore. When debugging anormal memory consumption, I saw that Conntrack now return a tuple:

```
In [5]: with Conntrack() as sock:
   ...:     print(type(sock.dump()))
   ...: 
<class 'tuple'>
```

It was a silent regression by us since we are iterating on it, as long as data was not too huge.

I had a look on the code, and it looks like this parameter is now ignored on NetlinkSocket class, the parent of Conntrack?

svinota:
Yep, that's a bug that I have overlooked. Sorry for that, will be fixed, ETA tomorrow. Do you need a release tag on it?

svinota:
Looks like problem is more complicated, so I have to take some more time. Still it is to be fixed with the highest priority.

svinota:
@ffourcot could you please take a look at this branch https://github.com/svinota/pyroute2/pull/1344 , if it helps? (it's not a complete solution, but I believe it should help a bit)

svinota:
Pls check if it is resolved now. Thanks!

ffourcot:
A big thank you, as said in the MR it is probably not CPU efficient (since we are calling a loop.run_until_complete at each iteration) but it fixes the RAM issue.

Do you have plan to add one AsyncNFCTSocket class?

svinota:
During this year I plan to provide asyncio interface to as many subsystems as I can :)

PR #1344, opened 2025-05-08
core: generator API for synchronous nlm_request

svinota:
Bug-Url: https://github.com/svinota/pyroute2/issues/1343


pyroute2/netlink/core.py
```diff
@@ -338,7 +356,9 @@ def event_loop(self):
                     message='Predefined event loop can not be used '
                     'in another thread',
                 )
-            self.local.event_loop = self.setup_event_loop()
+                self.local.event_loop = self.use_event_loop
```
ffourcot, line 359: I think that your are fixing a bug here, but it has side effect in our codebase:

```
   File "routing.py", line 2668, in _get_primary_interface_mac_address
     ipr.close()
   File "pyroute2/pyroute2/netlink/core.py", line 804, in close
     self.asyncore.close(code)
   File "pyroute2/pyroute2/netlink/core.py", line 473, in close
    self.event_loop.close()
   File "/usr/lib/python3.11/asyncio/unix_events.py", line 68, in close
     super().close()
   File "/usr/lib/python3.11/asyncio/selector_events.py", line 88, in close
     raise RuntimeError("Cannot close a running event loop")
```

So I was not able to look deeper today, but I will try to test this MR without this change tomorrow

svinota: Oh, that's a line from another fix, added here accidentally. And that's why it is breaking, the fix is not complete.

I removed the line, pushed over.

The issue that fixes this particular line will be addressed in the next PR for another bug.

ffourcot: Thank you, I tried the MR and it fixes our problem, thank you! It looks a little bit slower, but it does not consume all the RAM anymore

PR #1345, opened 2025-05-14
dhcp: decode bytes options as strings or lists to make them JSON encodable

etene:
This is a fix for an oversight on my part.

The DHCP client stores the whole ACK from the server in its leases.
That means when dumping a lease to a JSON file, all data in the ACK must be serializable.
However, i did not decode byte strings and these can't be JSON-encoded.

This PR changes the decoding logic so that byte strings in options are decoded as strings, and all bytes values are turned to a list of ints if undecodable.
This way, all options can be dumped to JSON successfully.


svinota:
Ok, will be merged tonight.

Thanks a lot!

PR #1348, opened 2025-05-22
common: introduce time.CLOCK_BOOTTIME to compute timeout

ffourcot:
time.time() is dangerous since in case of clock synchronization:
 * it can go backward and produce weird effects
 * it can go forward with a big jump

time.CLOCK_BOOTTIME provides safety for timeout checks.

See https://docs.python.org/3/library/time.html#time.CLOCK_BOOTTIME

One drawback: it is only available on Linux, after 2.6.39 version

### DIFF
--- a/pyroute2/common.py
+++ b/pyroute2/common.py
@@ -466,3 +466,14 @@ def msg_done(msg) -> bytes:
     # nlmsgerr struct alignment
     newmsg += b'\0' * 20
     return newmsg
+
+
+def get_boottime() -> float:
+    '''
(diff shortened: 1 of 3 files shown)


svinota:
Thanks!

svinota:
Just an idea — since we don’t need absolute time, only relative intervals, we could store the script’s start time as a constant in the config, for example, and use that, what do you think?

svinota:
@ffourcot another question — is there a reason to prefer here `time.CLOCK_BOOTTIME` here over cross-platform `time.monotonic()`?

PR #1350, opened 2025-05-27
nftables: missing await

rexzhang:


svinota:
Wow, how could I miss that. Thanks!

svinota:
@rexzhang thanks for contribution!

Issue #1381, opened 2025-07-26
Add support for conntrack zone


ArtsiomAntropau:
Hi, please add support for conntrack zone (CTA_ZONE) for filtration and ConntrackEntry

svinota:
Oke, planned for 0.9.5

(0.9.4 will be a fix release now on Monday)

ArtsiomAntropau:
can not make advanced tests, but looks good to me, thanks

svinota:
Then I close this ticket. Feel free to reopen or create a new one in the case of issues.

PR #1383, opened 2025-07-27
conntrack: add zone to dump and ConntrackEntry

svinota:
Bug-Url: https://github.com/svinota/pyroute2/issues/1381


svinota:
@ffourcot this change can possibly affect your code, so pls take a look before it's too late

PR #1387, opened 2025-08-03
ipset: async iterators

svinota:
Support iterators in listing methods

### DIFF
--- a/pyroute2/ipset.py
+++ b/pyroute2/ipset.py
@@ -21,7 +21,7 @@ from pyroute2.netlink import (
     NLM_F_REQUEST,
     NLMSG_ERROR,
 )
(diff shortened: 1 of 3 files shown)


svinota:
Attention @ffourcot , @inemajo , @orzel  this change can affect your code, pls take a look.

Short summary: `IPSet.list()` and `IPSet.headers()` are being changed to return an iterator instead of a tuple, thus making possible to perform lazy evaluation and avoid memory consumption in the case of huge numbers of ipset records.

orzel:
Ok, thanks for pointing out.
That looks like a good change anyway :)

ffourcot:
It looks good. I will try to provide "soon" one async class for WiSet. Our CI did not found any regression after the merge of this branch.

Issue #1364, opened 2025-07-14
NDB: Deleted IPs reappear if an error is encountered in future.


ekkagra:
```
# pip list | grep pyroute2
pyroute2                0.9.2
```

Say all IPs are deleted using del_ip()
```python

n = NDB()

>>> with n.interfaces["net1"] as net1:
...     net1.del_ip()
...
{'alt_ifname_list': [], 'state': 'up', 'ifname': 'net1', 'target': 'localhost', 'tflags': 0, 'index': 3, 'family': 0, 'ifi_type': 1, 'flags': 69699, 'change': 0, 'address': '66:37:0e:ff:1a:68', 'broadcast': 'ff:ff:ff:ff:ff:ff', 'mtu': 1500, 'link': 2, 'qdisc': 'noqueue', 'master': None, 'txqlen': 0, 'operstate': 'UP', 'linkmode': 0, 'net_ns_pid': None, 'ifalias': None, 'num_vf': None, 'group': 0, 'ext_mask': None, 'promiscuity': 0, 'num_tx_queues': 1, 'num_rx_queues': 1, 'carrier': 1, 'carrier_changes': 0, 'link_netnsid': 0, 'phys_port_name': None, 'proto_down': 0, 'gso_max_segs': 65535, 'gso_max_size': 65536, 'event': None, 'new_netnsid': None, 'if_netnsid': None, 'carrier_up_count': 0, 'carrier_down_count': 0, 'new_ifindex': None, 'min_mtu': 68, 'max_mtu': 65535, 'alt_ifname': None, 'parent_dev_name': None, 'parent_dev_bus_name': None, 'gro_max_size': None, 'tso_max_size': None, 'tso_max_segs': None, 'kind': 'macvlan', 'slave_kind': None, 'xdp_attached': None, 'xdp_flags': None, 'xdp_prog_id': None, 'xdp_drv_prog_id': None, 'xdp_skb_prog_id': None, 'xdp_hw_prog_id': None, 'macvlan_mode': 'bridge', 'macvlan_flags': 'none', 'macvlan_macaddr_mode': None, 'macvlan_macaddr': None, 'macvlan_macaddr_count': 0}

>>> n.interfaces["net1"].ipaddr.summary()

>>>
```

then again if some operation is called and it fails, the previously deleted IPs reappear
```python
# suppose del_ip is called again. it errors out, but previously deleted IPs reappear
>>> with n.interfaces["net1"] as net1:
...     net1.del_ip()
...

[...]

PR #1388, opened 2025-08-05
ndb: snapshot cleanup on create

svinota:
Bug-Url: https://github.com/svinota/pyroute2/issues/1364


ekkagra:
I installed pyroute2 based on this PR and see that now
1. future errors don't bring back previous state of addrs on interface
2. del_ip() doesn't error when no addrs were present

Looks good 👍

PR #1405, opened 2025-08-21
Async wiset

ffourcot:
Hello,

As promised, a first version for AsyncWiSet support. It works more a less like WiSet, but with less "magic" and more exceptions:
- Netlink sockets are not opened without context manager or explicit opening
- In the same spirit, "content" property raise explicit exception instead of reloading content when it was not loaded before

Do not hesitate if you have any comment or idea to improve it.


svinota:
@ffourcot pls take a look: https://github.com/svinota/pyroute2/pull/1406

PR #1411, opened 2025-08-27
Generalize exception_factory and add WiRouting class

ffourcot:
Hello @svinota,

We are working to replace our custom code (based in IPRoute) to move to AsyncIPRoute. We think that it could be valuable to share this, and to integrate it in pyroute2. The main idea is to provide one high-level API to manage network, without to manage low level details of netlink (for example, nobody care that an interface rename is a "set" on ifname).

It's a proof of concept. Let me know if you are interested, we will then propose more pull requests to extend features. Basically it could provide, on top of AsyncIPRoute:

- Custom exception. We think that parsing error code on each call is time consuming;
- Return objects as python dataclass (not yet implemented)
- As recommended in kernel documentation, enable ext_act and strict_check by default

This MR will change expected signature of exception_factory, I hope that it is ok.


pyroute2/netlink/__init__.py
```diff
@@ -2447,6 +2448,14 @@ class nlmsgerr(nlmsg):
         ('NLMSGERR_ATTR_COOKIE', 'uint8'),
     )
 
+    @property
+    def orig_msg(self) -> Dict:
```
svinota, line 2452: Yep, a good thing to have, thanks

pyroute2/wirouting.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+def exception_factory(err, msg):
+    try:
+        return exception_map[msg.orig_type][err.code](*err.args)
+    except LookupError:
+        return err
+
+
+class WiRoute(AsyncIPRoute):
```
svinota, line 54: Do we plan the synchronous counterpart?

tests/test_linux/test_wiroute.py
```diff
@@ -0,0 +1,50 @@
+import pytest
+
+from pyroute2.wirouting import InterfaceDoesNotExist, WiRoute, InterfaceExists
+
+pytestmark = [pytest.mark.asyncio]
+BAD_INTERFACE_NAME = "doesnotexist"
+
+
+async def test_specialized_exceptions():
```
svinota, line 9: if possible, pls mark async tests `@pytest.mark.asyncio`

ffourcot: See some lines above, the mark is done for all the file:

```
pytestmark = [pytest.mark.asyncio]
```

pyroute2/wirouting.py
ffourcot: Not really. It's a good think to provide synchronous code as compatibility for previous code already deployed, but I think that for new code it's better to let caller do loop.run_until_complete()

svinota:
I believe it's ok to proceed with the branch, @ffourcot , and probably I could reuse some ideas in NDB to provide compatible formats

ffourcot:
@svinota ok great. So in that case go to merge for me, we will add features in future merge requests step-by-step. Thank you

ffourcot:
@svinota hello, I think that I fixed the CI (sorry for the noise, I was not very efficient and I pushed several times)

svinota:
Reading the code

pyroute2/wirouting.py
```diff
@@ -0,0 +1,72 @@
...
+    def __init__(self, *args, **kwargs):
+        for key in ("ext_ack", "strict_check"):
+            kwargs.setdefault(key, True)
```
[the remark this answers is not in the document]
ffourcot: Would you like to set it to False in that case? My goal here was to not overwrite an explicit ```False```. I'm note sure that ```None``` is an acceptable value, since both keys are defined as boolean, but I can update the code to accept it.

PR #1412, opened 2025-08-28
dhcp: add support for encoding & decoding option 118 (subnet selection)

etene:
Hi, it's me again !

I've been working on a reproducer for an obscure dnsmasq issue, and I needed to be able to send the subnet selection option. It is silently discarded at the moment since there is no mapping for it in `dhcp4msg.options`.

It's a very small MR, i just added an tested this single option.

(see [RFC3011](https://www.rfc-editor.org/rfc/rfc3011.html) for details)


etene:
Thanks for merging so fast :)

PR #1416, opened 2025-09-15
wiset: add delete() method and fix add method for ipset without IP address

ffourcot:
I implemented add() before but without method to remove an entry.


ffourcot:
@svinota since the both wiset related merge requests we short, I regrouped it. It's not a lot of code but both commits are mandatory to migrate our internal code to AsyncWiSet

PR #1422, opened 2025-09-16
wiset/async: fix etype argument on add function

ffourcot:
etype was set on sync method, but not the async one. It was impossible to add complex objets (not IP or with arguments like skbmark)


ffourcot:
Merged with https://github.com/svinota/pyroute2/pull/1416

PR #1459, opened 2026-01-16
ipvs: add support for fwmarks in service keys

apollo13:
services working on a fwmark weren't able to dump real servers since the fwmark was missing from the service key.


svinota:
Thanks, merging!

PR #1468, opened 2026-03-20
tc u32: support mark parameter

liam-mcb:
Support setting TCA_U32_MARK parameter to filter by fwmark. Allow keys parameter to be empty.


svinota:
Thanks!

PR #1475, opened 2026-06-01
rtnl/ndmsg: add missing NTF flags

rouven0:
This PR completes the list of available NTF flags.
I came across a situation in which I needed the `NFT_STICKY` flag for a bunch of static fdb entries and found the functionality missing.

See also: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/uapi/linux/neighbour.h#n51


svinota:
Thanks!

PR #1476, opened 2026-06-05
Wiroute link features

inemajo:
Hi svinota, here’s a PR regarding wirouting (the base implementation for ip-route and ip-link) in read-only + view mode.

We plan to migrate as much of our homemade custom code as possible to wirouting from pyroute2. If you have any suggestions or comments, feel free to share!


svinota:
What are Python version requirements for wirouting, so I would guard it in the tests?

inemajo:
I’ve only tested it with Python 3.13, but I can test other versions if needed.

svinota:
Something not working in the tests:
```
E   TypeError: Too few arguments for typing.AsyncGenerator; actual 1, expected 2
```
See more in the logs: https://github.com/svinota/pyroute2/actions/runs/27016792965/job/79985687570?pr=1476

svinota:
The CI issue is caused by the new worker, so don't mind it. Otherwise LGTM, merging.

PR #1483, opened 2026-06-30
Ethtool off flags

inemajo:
Hello @svinota, recently we needed the "generic-receive-offload" flag/feature in ethtool, but it is located in a different ioctl than the current get_feature. Here is the implementation to retrieve the off_flags.


svinota:
Thanks, merging!

